PR #40, opened 2023-04-07
Feature/scipy1.10

zebengberg:
#### Features

- Improve interpolation performance by cythonizing linear interpolation. This extends the approach taken in [scipy 1.10](https://github.com/scipy/scipy/pull/17291). The pycontrails [cython routines](pycontrails/core/rgi_cython.pyx) allow for both float64 and float32 grids via cython fused types (the current scipy implementation assumes float64). In addition, interpolation up to dimension 4 is supported (the scipy implementation supports dimension 1 and 2).
- Officially support [scipy 1.10](https://docs.scipy.org/doc/scipy/index.html).

#### Breaking changes

- Change `MetDataset` and `MetDataArray` conventions: underlying dimension coordinates are automatically promoted to float64.
- Change how datetime arrays are converted to floating values for interpolation. The new approach introduces small differences compared with the previous implementation. These differences are significant enough to see relative differences in CoCiP predictions on the order of 1e-4.

#### Fixes

- Unit tests no longer raise errors when the `pycontrails-bada` package is not installed. Instead, some tests are skipped.

#### Internals

- Make the `interpolation` module more aligned with [scipy 1.10 enhancements](https://docs.scipy.org/doc/scipy/release.1.10.0.html#scipy-interpolate-improvements) to the `RegularGridInterpolator`. In particular, grid coordinates now must be float64.
- Use [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) to build wheels for Linux, macOS (arm64 and x86_64), and Windows on [release](.github/workflows/release.yaml) in Github Actions. Allow this workflow to be triggered manually to test the release process without actually publishing to PyPI.
- Simplify interpolation with pre-computed indices (invoked with the model parameter `interpolation_use_indices`) via a `RGIArtifacts` interface.
- Slight performance enhancements to the `met` module.

## Tests

- [x] CI tests pass

## Reviewer

@mlshapiro I still have work to do on the distribution side of this. I may want to hold off merging until I can test better on windows.

## Interpolation performance

I've pasted plots of interpolation times on the head main and the head of this branch.

Both axes are log-scaled.

Screenshot 2023-04-07 at 12 22 55 PM

Both axes are linear scaled.

Screenshot 2023-04-07 at 12 23 31 PM

The same information in table form. A larger ratio is better.

|   vector_size |   ratio |
|--------------:|--------:|
|            10 |    2.33 |
|            12 |    3.22 |
|            15 |    2.26 |
|            20 |    2.28 |
|            25 |    2.35 |
|            32 |    2.27 |
|            40 |    2.28 |
|            51 |    2.3  |
|            65 |    2.3  |
|            82 |    2.3  |
|           104 |    2.34 |
|           132 |    3.19 |
|           167 |    3.04 |
|           212 |    3.12 |
|           268 |    3.15 |
|           339 |    3.09 |
|           429 |    3.06 |
|           542 |    2.95 |
|           686 |    2.54 |
|           868 |    2.6  |
|          1098 |    2.54 |
|          1389 |    2.33 |
|          1757 |    2.16 |
|          2222 |    2.08 |
|          2811 |    1.97 |
|          3556 |    1.93 |
|          4498 |    1.89 |
|          5689 |    1.87 |
|          7196 |    1.87 |
|          9102 |    1.86 |
|         11513 |    1.79 |
|         14563 |    1.82 |
|         18420 |    1.78 |
|         23299 |    1.84 |
|         29470 |    1.86 |
|         37275 |    2.03 |
|         47148 |    2    |
|         59636 |    2.02 |
|         75431 |    2.04 |
|         95409 |    2.03 |
|        120679 |    2.06 |
|        152641 |    2.04 |
|        193069 |    2.11 |
|        244205 |    2.14 |
|        308884 |    2.24 |
|        390693 |    2.28 |
|        494171 |    2.16 |
|        625055 |    2.19 |
|        790604 |    2.15 |
|       1000000 |    2.28 |

### DIFF
--- a/pycontrails/core/met.py
+++ b/pycontrails/core/met.py
@@ -1229,7 +1229,7 @@ class MetDataArray(MetBase):
     ...     "longitude": np.arange(-20, 20),
     ...     "latitude": np.arange(-30, 30),
     ...     "level": [220, 240, 260, 280],
-    ...     "time": [np.datetime64("2021-08-01T12"), np.datetime64("2021-08-01T16")]
+    ...     "time": [np.datetime64("2021-08-01T12", "ns"), np.datetime64("2021-08-01T16", "ns")]
     ...     }
     >>> da = xr.DataArray(rng.random((40, 60, 4, 2)), dims=coords.keys(), coords=coords)
 
--- a/pycontrails/core/met.py
+++ b/pycontrails/core/met.py
@@ -27,6 +27,7 @@ from typing import (
 )
 
 import numpy as np
+import numpy.typing as npt
 import xarray as xr
 from overrides import overrides
 
@@ -46,6 +47,8 @@ if TYPE_CHECKING:
 XArrayType = TypeVar("XArrayType", xr.Dataset, xr.DataArray)
 MetDataType = TypeVar("MetDataType", "MetDataset", "MetDataArray")
 DatasetType = TypeVar("DatasetType", xr.Dataset, "MetDataset")
+
+COORD_DTYPE = np.float64
 
 
 class MetBase(ABC, Generic[XArrayType]):
@@ -104,31 +107,38 @@ class MetBase(ABC, Generic[XArrayType]):
             If longitude values are not contained in the interval [-180, 180].
         """
         longitude = self.variables["longitude"].values
+        if longitude.dtype != COORD_DTYPE:
+            raise ValueError(
+                "Longitude values must be of type float64. "
+                "Initiate with 'copy=True' to convert to float64. "
+                "Initiate with 'validate=False' to skip validation."
+            )
 
         if self.is_wrapped:
             # Relax verification if the longitude has already been processed and wrapped
-            if longitude[-1] > 360:
+            if longitude[-1] > 360.0:
                 raise ValueError(
                     "Longitude contains values > 360. Shift to WGS84 with "
                     "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
                 )
-            if longitude[0] < -360:
+            if longitude[0] < -360.0:
                 raise ValueError(
                     "Longitude contains values < -360. Shift to WGS84 with "
                     "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
                 )
-        else:
-            # Strict!
-            if longitude[-1] > 180:
-                raise ValueError(
-                    "Longitude contains values > 180. Shift to WGS84 with "
-                    "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
-                )
-            if longitude[0] < -180:
-                raise ValueError(
-                    "Longitude contains values < -180. Shift to WGS84 with "
-                    "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
-                )
+            return
+
+        # Strict!
+        if longitude[-1] > 180.0:
+            raise ValueError(
+                "Longitude contains values > 180. Shift to WGS84 with "
+                "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
+            )
+        if longitude[0] < -180.0:
+            raise ValueError(
+                "Longitude contains values < -180. Shift to WGS84 with "
+                "'data.assign_coords(longitude=(((data.longitude + 180) % 360) - 180))'"
+            )
 
     def _validate_latitude(self) -> None:
         """Check latitude bounds.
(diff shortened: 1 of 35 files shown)


mlshapiro:
Wow these numbers are pretty astounding. I agree that we should wait to test on Windows before merging, but amazing work getting this packaged for linux/mac.

I'll review test more deeply (e.g. benchmark tests)

pycontrails/core/met.py
```diff
@@ -256,29 +284,47 @@ def _preprocess_dims(self, wrap_longitude: bool) -> None:
         if self.is_single_level:
             # add level attributes to reflect surface level
```
mlshapiro, line 285: small thing - `surface` should be `single` in comment here

pycontrails/core/met.py
```diff
@@ -256,29 +284,47 @@ def _preprocess_dims(self, wrap_longitude: bool) -> None:
...
+
+        # add pressure level attributes
+        self.data["level"].attrs.update(units="hPa", long_name="Pressure", positive="down")
+
+        # add altitude and air_pressure
+
+        # XXX: use the dtype of the data to determine the precision of these coordinates
+        # There are two competing conventions here:
+        # - coordinate data should be float64
+        # - gridded data is typically float32
```
mlshapiro, line 300: I would call this `data values` or `gridded data values`

pycontrails/core/met.py
```diff
@@ -970,20 +1016,20 @@ def to_vector(self, transfer_attrs: bool = True) -> vector_module.GeoVectorDatas
     @classmethod
     def from_coords(
         cls,
-        longitude: np.ndarray | list[float] | float,
-        latitude: np.ndarray | list[float] | float,
-        level: np.ndarray | list[float] | float,
-        time: np.ndarray | list[np.datetime64] | np.datetime64,
+        longitude: npt.ArrayLike | float,
```
mlshapiro, line 1019: oh awesome! npt.ArrayLike!

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -448,6 +450,9 @@ def _generate_new_vectors(
 
                 # Convert the 4D grid to a vector
                 vector = source_slice.to_vector()
+                vector.update(longitude=vector["longitude"].astype(self._target_dtype, copy=False))
```
mlshapiro, line 453: So if I interpret this right - you are keeping vector dtypes the same as the met data values?  And the reason to have a different standard here is to reduce memory overhead?

setup.py
mlshapiro, line 1: Too bad this isn't toml'd yet

tests/unit/test_cache.py
```diff
@@ -55,6 +57,7 @@ def test_cache_bad_path(self) -> None:
         disk_path = "".join(random.choice(["x", "y"]) for _ in range(20))
         assert not _cache.exists(disk_path)
 
+    @pytest.mark.skipif(not pathlib.Path("README.md").is_file(), reason="No README.md")
```
mlshapiro, line 60: Where does this come from?

tests/unit/test_cocip.py
```diff
@@ -929,7 +936,6 @@ def test_radiative_heating_effects_param(fl: Flight, met: MetDataset, rad: MetDa
         "dt_integration": np.timedelta64(10, "m"),
         "process_emissions": False,
         "humidity_scaling": ExponentialBoostLatitudeCorrectionHumidityScaling(),
-        "interpolation_bounds_error": True,
```
mlshapiro, line 932: Did we get rid of this parameter? I don't see where this is changed in the code elsewhere

mlshapiro (approved):
Took another pass. This looks great.

pycontrails/models/cocipgrid/cocip_grid.py
zebengberg: Yeah -- you got it right. Sorry this so opaque.

Now, we automatically convert `MetDataset` coords to float64. But I want to be able to run `CocipGrid` with float32 precision for memory and performance reasons. This is a cheap way to let the underlying grid dtype dictate the precision when we convert to vector here.

Definitely confusing!

tests/unit/test_cache.py
zebengberg: The test itself does funny stuff with the README file (uploading it to the cloud, etc). We're implicitly assuming that we're running the tests from the root of the pycontrails directory (which isn't always the case).

tests/unit/test_cocip.py
zebengberg: The parameter is still there, but there was some sort of bug in which an error wasn't raised despite contrails evolving out of bounds. It was somewhat tangled up with the scipy 1.9 implementation. The scipy 1.10 `RegularGridInterpolator` has changed a lot.

PR #57, opened 2023-05-17
Piecewise Linear Fitting

trdean1:
## Changes

> Adds two methods to the `Flight` module. One is a class method to the class `Flight`. The other operates on an array of times and altitudes

#### Internals

Calls the library `pwlf` to fit a series of line segments to a flight profile.  A Savitzky-Golay smoothing filter is then applied

## Tests

- [ x] QC test passes locally (`make test`)
- [ ] CI tests pass

## Reviewer

> @mlshapiro 
> @zebengberg

### DIFF
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -942,10 +942,13 @@ class Flight(GeoVectorDataset):
         Flight
             Smoothed flight
         """
-        elapsed_time = [0] + np.nancumsum(self.segment_duration())
+        # np.roll pushes the last NaN value from `segment_duration` to the front
+        # so the elapsed time at the first waypoint will be 0
+        seg_dur = self.segment_duration(dtype=np.float64)
+        elapsed_time = np.nancumsum(np.roll(seg_dur, 1))
         alt_ft = fit_altitude(
             elapsed_time,
-            self.altitude_ft,
+            np.copy(self.altitude_ft),
             max_segments,
             pop,
             r2_target,
@@ -1750,6 +1753,7 @@ def fit_altitude(
     lvl = np.round(m2.intercepts[mask], -3)
     time_stack = np.repeat(elapsed_time[:, np.newaxis], lvl.size, axis=1)
     filt = (time_stack >= bounds[0]) & (time_stack <= bounds[1])
+    altitude_ft = np.copy(altitude_ft)
     for i in range(lvl.size):
         altitude_ft[filt[:, i]] = lvl[i]
 
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
+import pwlf
 import pyproj
 import scipy.signal
 from overrides import overrides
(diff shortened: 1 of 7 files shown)


pycontrails/core/flight.py
```diff
@@ -11,6 +11,7 @@
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
+import pwlf
```
zebengberg, line 14: I would suggest only requiring this as an optional dependency. So do the import inside of the `fit_altitude` function. When we do this thing, we try to give a nice error message. So something like:

```python
try:
   import pwlf
except ModuleNotFoundError as e:
   raise ModuleNotFoundError("The 'fit_altitude' function requires the 'pwlf' package. This can be installed with ...") from e
```

pycontrails/core/flight.py
```diff
@@ -900,6 +901,62 @@ def resample_and_fill(
         df = df.reset_index()
         return Flight(data=df, attrs=self.attrs)
 
+    def fit_flight_profile(
```
zebengberg, line 904: To be safe, I think you should add this same method to the `Fleet` class, and it can simply raise `NotImplementedError("Only implemented for Flight instances")`

pycontrails/core/flight.py
```diff
@@ -900,6 +901,62 @@ def resample_and_fill(
...
+            Parameter for :func:`scipy.signal.savgol_filter`
+        sg_polyorder: int, optional
+            Parameter for :func:`scipy.signal.savgol_filter`
+
+        Returns
+        -------
+        Flight
+            Smoothed flight
+        """
+        elapsed_time = [0] + np.nancumsum(self.segment_duration())
```
zebengberg, line 945: What's the `[0]` doing?

pycontrails/core/flight.py
```diff
@@ -900,6 +901,62 @@ def resample_and_fill(
...
+        alt_ft = fit_altitude(
+            elapsed_time,
+            self["altitude_ft"],
```
zebengberg, line 948: `self["altitude_ft"]` -> `self.altitude_ft`

pycontrails/core/flight.py
```diff
@@ -1626,3 +1683,69 @@ def segment_rocd(
...
     return out
+
+
+def fit_altitude(
+    elapsed_time: npt.NDArray[np.float_],
+    altitude_ft: npt.NDArray[np.float_],
+    max_segments: int = 30,
+    pop: int = 3,
+    r2_target: float = 0.999,
+    max_cruise_rocd: float = 10,
```
zebengberg, line 1694: 10 -> 10.0

pycontrails/core/flight.py
```diff
@@ -1626,3 +1683,69 @@ def segment_rocd(
...
+    for i in range(1, max_segments):
+        m2 = pwlf.PiecewiseLinFit(elapsed_time, altitude_ft)
+        r = m2.fitfast(i, pop)
+        r2 = m2.r_squared()
+        if r2 > r2_target:
+            break
+
+    mask = abs(m2.slopes) < max_cruise_rocd / 60.0
+    bounds = r[:-1][mask], r[1:][mask]
+    lvl = np.round(m2.intercepts[mask] / 1000) * 1000
```
zebengberg, line 1743: Does this work?

```
lvl = np.round(m2.intercepts[mask], -3)
```

pycontrails/core/flight.py
```diff
@@ -1626,3 +1683,69 @@ def segment_rocd(
...
+    for i in range(len(bounds[0])):
+        altitude_ft[
+            np.bitwise_and(elapsed_time >= float(bounds[0][i]), elapsed_time <= float(bounds[1][i]))
+        ] = lvl[i]
```
zebengberg, line 1747: Can this be vectorized? I may not be understanding it ... would something like this work:

```
filt = (elapsed_time >= bounds[0]) & (elapsed_time <= bounds[1])
altitude_ft[filt] = lvl[filt]
```
?

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "numpy>=1.22",
     "overrides>=6.1",
     "pandas>=1.4",
+    "pwlf>=2.2.1",
```
zebengberg, line 38: I'd suggest making a new custom group in `[project.optional-dependencies]`

tests/unit/test_flight.py
```diff
@@ -220,6 +222,18 @@ def test_flight_validate(flight_data: pd.DataFrame) -> None:
         Flight(df)
 
 
+def test_flight_fitting() -> None:
+    ds = pd.read_parquet(get_static_path("flight-spire-data-cleaning.pq"))
+    ds.rename(columns={"altitude_baro": "altitude_ft", "timestamp": "time"}, inplace=True)
+    ds["time"] = ds["time"].dt.tz_localize(None)
+    flight = Flight(ds[ds["callsign"] == "BAW506"], drop_duplicated_times=True)
+    smoothed_flight = flight.fit_flight_profile()
```
zebengberg, line 230: Could you add an assert here that the original `flight` isn't modified by this? IDK the best way to do this -- you could pin the mean before you call the `fit_flight_profile` method, and check that after you call it, the mean hasn't changed.

pycontrails/core/flight.py
zebengberg: (A flight isn't guaranteed to have a "altitude_ft" column)

zebengberg (approved):
Nice Tom!!!

zebengberg:
> I'm guessing we might want to tweak the names of these methods and where the test is placed. Let me know and I can change it. Wanted to get something up for review.

I think the test is in the right place. I would suggest renaming the `Flight` method to `fit_altitude` as well.

pycontrails/core/flight.py
trdean1: That doesn't work, but this does, not sure if it's much better:

```
  bounds = r[:-1][mask], r[1:][mask]
  lvl = np.round(m2.intercepts[mask], -3)
  time_stack = np.repeat(elapsed_time[:,np.newaxis],lvl.size,axis=1)
  filt = (time_stack >= bounds[0]) & (time_stack <= bounds[1])
  for i in range(lvl.size):
      altitude_ft[filt[:,i]] = lvl[i]
```

pycontrails/core/flight.py
trdean1: The first waypoint is at time 0, and using `np.nancumsum` causes the last NaN value to get dropped

PR #58, opened 2023-05-22
Feature/ps model2

### LINKED ISSUE #36: Remove Aircraft class and document special Flight.attr keys
**Description**

I struggle to understand the difference between `engine_uid` and `engine_name`. According to the documentation, the `Aircraft` class contains the attribute `engine_name`. The `Flight` class contains nothing related to the engine itself, only an unspecified `attrs` attribute. 

https://py.contrails.earth/api/pycontrails.Aircraft.html
https://py.contrails.earth/api/pycontrails.Flight.html

Internally, I find that `engine_uid` is used for BADA4 support. 

https://github.com/contrailcirrus/pycontrails-bada/blob/32ce4a5544929fcf78f490bc353ae74c965929d9/pycontrails_bada/bada_interface.py#L124

In `BADAFlight.eval` the docstring says

```
If the following variables are not provided in the flight attribute, the default aircraft-engine assignment and
aircraft mass properties from BADA will be assumed:
- engine_uid
```

https://github.com/contrailcirrus/pycontrails-bada/blob/4968c976cd2b35efc94b134f49d7c201c6f60df2/pycontrails_bada/bada_model.py#L396

So, `engine_uid` should be given in the `Flight.attrs` dictionary? What is `Aircraft.engine_name` used for? Maybe some documentation is lacking on the usage and the differences?

zebengberg:
Closes #36 #11

### Features

- Implement a basic working version of the [Poll-Schumann (PS) aircraft performance model](https://doi.org/10.1017/aer.2020.62). This is experimental and may undergo revision in future releases. The PS Model currently supports the following 53 aircraft types:

  - A30B
  - A306
  - A310
  - A313
  - A318
  - A319
  - A320
  - A321
  - A332
  - A333
  - A342
  - A343
  - A345
  - A346
  - A359
  - A388
  - B712
  - B732
  - B733
  - B734
  - B735
  - B736
  - B737
  - B738
  - B739
  - B742
  - B743
  - B744
  - B748
  - B752
  - B753
  - B762
  - B763
  - B764
  - B77L
  - B772
  - B77W
  - B773
  - B788
  - B789
  - E135
  - E145
  - E170
  - E195
  - MD82
  - MD83
  - GLF5
  - CRJ9
  - DC93
  - RJ1H
  - B722
  - A20N
  - A21N

- Improve the runtime of instantiating the `Emissions` model by a factor of 10-15x. This translates to a time savings of several hundred milliseconds on modern hardware. This improvement is achieved by more efficient parsing of the underlying static data and by deferring the construction of the interpolation artifacts until they are needed.
- Automatically use a default engine type from the aircraft type in the `Emissions` model if an `engine_uid` parameter is not included on the `source`. This itself is configurable via the `use_default_engine_uid` parameter on the `Emissions` model. The default mappings from aircraft types to engines is included in `pycontrails/models/emissions/static/default-engine-uids.csv`.

### Breaking changes

- Remove the `Aircraft` dataclass from `Flight` instantiation. Any code previously using this should instead directly pass additional `attrs` to the `Flight` constructor.

### Fixes

- Remove the `fuel` parameter from the `Emissions` model. This is inferred directly from the `source` parameter in `Emissions.eval`.

### Internals

- Use `ruff` in place of `pydocstyle` for linting docstrings.
- Update the `AircraftPerformance` template based on the patterns used in the new `PSModel` class. This may change again in the future.

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer


Makefile
```diff
@@ -93,7 +90,7 @@ pytest-cov:
 		--ignore=tests/unit/test_zarr.py \
 		tests/unit
 
-test: ruff mypy black-check nb-black-check pydocstyle pytest doctest nb-test
```
mlshapiro, line 96: 👍

pycontrails/models/emissions/static/default-engine-uids.csv
mlshapiro, line 1: We should probably leave a note to make an `Emissions` tutorial that surfaces some of these static files.

pycontrails/physics/jet.py
```diff
@@ -275,6 +380,19 @@ def initial_aircraft_mass(
 ) -> float:
     """Estimate initial aircraft mass as a function of load factor and fuel requirements.
 
+    This function uses the following equation::
+
+        TOM = OEM + PM + FM_nc + TFM
+            = OEM + LF * MPM + FM_nc + TFM
+
+    where:
```
mlshapiro, line 388: What is `LF`?

pycontrails/physics/jet.py
```diff
@@ -24,14 +24,62 @@
 # -------------------
 
 
+def acceleration(
+    true_airspeed: npt.NDArray[np.float_], dt: npt.NDArray[np.float_]
```
mlshapiro, line 28: We call this `segment_duration` in `jet.fuel_burn`. I think I prefer the verbose name.

```
    segment_duration: npt.NDArray[np.float_]
        Time difference between waypoints, [:math:`s`]
```

We should add a `See Also` to both of these for `flight.segment_duration`
```
    See Also
    --------
    :func:`flight.segment_duration`

```

pycontrails/physics/jet.py
```diff
@@ -24,14 +24,62 @@
...
+    """
+    dv_dt = np.empty_like(true_airspeed)
+    dv_dt[:-1] = np.diff(true_airspeed) / dt[:-1]
+    dv_dt[-1] = 0.0
+    np.nan_to_num(dv_dt, copy=False)
+    return dv_dt
+
+
+def climb_descent_angle(
+    true_airspeed: npt.NDArray[np.float_], rocd_ms: npt.NDArray[np.float_]
```
mlshapiro, line 52: Are we actually using rocd in meters / s?  I found when I refactored this last time that we converted back and forth multiple times, so I actually stuck with `rocd` in `feet / min`. See e.g. `jet.reserve_fuel_requirements`.  

If you really get rocd in meters / second and don't re-convert it elsewhere, then I think its fine to have both types of units.

Ideally we'd keep all these units the same. If you go back to `ft / min`, then I would add a "See Also" for `flight.segment_rocd`:

```
    See Also
    --------
    :func:`flight.segment_rocd`
    :func:`flight.segment_true_airspeed`
```

pycontrails/physics/units.py
```diff
@@ -56,17 +57,17 @@ def ft_to_pl(h: ArrayScalarLike) -> ArrayScalarLike:
     return m_to_pl(ft_to_m(h))
 
 
-def kelvin_to_celcius(kelvin: np.ndarray) -> np.ndarray:
+def kelvin_to_celcius(kelvin: ArrayScalarLike) -> ArrayScalarLike:
```
mlshapiro, line 60: *Celcius is spelled wrong here --> `celsius`

pycontrails/models/aircraft_performance.py
```diff
@@ -4,34 +4,405 @@
...
+    n_iter: int = 5
 
 
 class AircraftPerformance(Model):
     """
     Support for standardizing aircraft performance methodologies.
 
-    Currently just a container until additional models are implemented.
+    This class provides a :meth:`simulate_fuel_and_performance` method for
+    iteratively calculating aircraft mass and fuel flow rate. The implementing
```
mlshapiro, line 42: I'd add a new line before `. The implementing...`

pycontrails/models/aircraft_performance.py
```diff
@@ -4,34 +4,405 @@
...
+        **kwargs : Any
+            Additional keyword arguments to pass to the model.
+
+        Returns
+        -------
+        AircraftPerformanceData
+            Derived performance metrics at each waypoint.
+        """
+
+    def ensure_tas_on_source(self) -> None:
```
mlshapiro, line 383: I find the `tas` label is not universal, so I'd probably make this way verbose as `ensure_true_airspeed` or `ensure_true_airspeed_on_source`

mlshapiro (approved):
Wow, so much great cleanup and abstraction in here! The `AircraftPerformance` model works, although it may be worth documenting how this model gets implemented a little differently than others (e.g. `eval` would rely on `simulate_...` which in turn calls `calculate_...`, or perhaps you just skip the simulate step, etc)

Just a few little doc/typos!

I didn't review emissions closely, but I'm glad to see it got a refactor!

pycontrails/physics/jet.py
zebengberg: thanks!

a29ce05

pycontrails/physics/jet.py
zebengberg: 9a57e5e

pycontrails/models/aircraft_performance.py
zebengberg: 2c70513

pycontrails/physics/units.py
zebengberg: 🤦 ac34d70

pycontrails/physics/jet.py
zebengberg: We do need rocd in m/s because the unit must agree with `true_airspeed`. But you're absolutely right that having several ROCDs is confusing. Because we don't use `rocd_ms` elsewhere, I'll make the conversion internal to this function.

zebengberg: ae912d7

zebengberg:
I tried to better document: e4f68aa

pycontrails/models/aircraft_performance.py
zebengberg: e4f68aa

pycontrails/models/emissions/static/default-engine-uids.csv
zebengberg: Definitely ... and we should try to introduce the new `PSModel` into some of our tutorials as well.

PR #62, opened 2023-06-07
Feature/hs refactor

### LINKED ISSUE #61: Determine appropriate interpolation methodology for pressure level and specific humidity
**Description**

In discussions with collaborators, we've received the suggestion to interpolate pressure levels on a log scale to better approximate geometric altitude.

We've also heard the recommendation to interpolate specific humidity as the log(specific humidity) to account for the wide range of value.

**Alternatives**

- Linear interpolation (current method)

**Additional context**

Per discussion 

@zebengberg @mccloskey

zebengberg:
Closes #61 

### Features

- Add new experimental `interpolation_q_method` field to the `ModelParams` data class. This parameter controls the interpolation methodology when interpolation against gridded specific humidity. The possible values are:
  - `"linear-q"`: Interpolate linearly against specific humidity. This is the default behavior and is the same as the previous behavior.
  - `"cubic-spline"`: Apply cubic-spline scaling to the interpolation table vertical coordinate before interpolating linearly against specific humidity.
  - `"log-q-log-p"`: Interpolate in the log-log domain against specific humidity and pressure.

  This interpolation parameter is used when calling `pycontrails.core.models.interpolate_met`. It can also be used directly with the new lower-level `pycontrails.core.models.interpolate_gridded_specific_humidity` function.
- Add new experimental `HistogramMatching` humidity scaling model to match RHi values against IAGOS observations. The previous `HistogramMatchingWithEckel` scaling is still available when working with ERA5 ensemble members.

### Breaking changes

- Add an optional `q_method` parameter to the `pycontrails.core.models.interpolate_met` function. The default value `"linear-q"` agrees with the previous behavior.
- Change function signatures in the `cocip.py` module for consistency. The `interp_kwargs` parameter is now unpacked in the `calc_timestep_meterology` signature. Rename `kwargs` to `interp_kwargs` where appropriate.

### Fixes

- Recompute and extend quantiles for histogram matching humidity scaling. Quantiles are now available for each combination of `q_method` and the following ERA5 data products: reanalysis and ensemble members 0-9. This data is available as a parquet file and is packaged with `pycontrails`.
- Fix the precomputed Eckel coefficients. Previous values where computed for different interpolation assumptions and were not correct for the default interpolation method.
- Clip the scaled humidity values computed by the `humidity_scaling.eckel_scaling` function to ensure that they are non-negative. Previously, both relative and specific humidity values arising from Eckel scaling could be negative.
- Avoid extraneous copy when calling `VectorDataset.sum`.

### Internals

- Refactor auxillary functions used by `HistogramMatchingWithEckel` to better isolated histogram matching from Eckel scaling.
- Refactor `intersect_met` method in `pycontrails.core.models` to handle experimental `q_method` parameter.
- Include precomputed humidity lapse rate values in the new `pycontrails.core.models._load_spline` function.

## Tests

- [x] CI tests pass

## Reviewer

@mlshapiro 
@mccloskey


pycontrails/core/models.py
```diff
@@ -311,14 +324,18 @@ def eval(self, source: Any = None, **params: Any) -> ModelOutput:
...
+
+            - "method"
+            - "bounds_error"
+            - "fill_value"
+            - "localize"
+            - "use_indices"
+
+            as determined by :attr:`params`.
         """
         return {
```
mlshapiro, line 340: It feels like this should contain the `interpolation_q_method` and then you could remove the `interpolation_` prefix?

pycontrails/core/models.py
```diff
@@ -553,11 +570,21 @@ def set_source_met(
             # take the first var name output from ensure_vars
             met_key = self.met.ensure_vars(var)[0]
 
+            q_method = self.params["interpolation_q_method"]
```
mlshapiro, line 573: Would be unnecessary if you put this value in `interp_kwargs`

pycontrails/core/models.py
```diff
@@ -553,11 +570,21 @@ def set_source_met(
...
+
             # interpolate GeoVectorDataset
             if isinstance(self.source, GeoVectorDataset):
-                interpolate_met(self.met, self.source, met_key, **self.interp_kwargs)
+                interpolate_met(
+                    self.met, self.source, met_key, q_method=q_method, **self.interp_kwargs
```
mlshapiro, line 578: You could remove `q_method` from this interface if it were in `interp_kwargs`

pycontrails/core/models.py
```diff
@@ -72,6 +75,16 @@ class ModelParams:
...
 
+    #: Experimental. Alternative interpolation method to account for specific humidity
+    #: lapse rate bias. Must be one of "linear-q", "cubic-spline", or "log-q-log-p".
+    #: The "linear-q" method applies naive linear interpolation over gridded met data.
+    #: The "cubic-spline" method applies a custom stretching of the met interpolation
+    #: table to account for the specific humidity lapse rate bias. The "log-q-log-p"
+    #: method interpolates in the log of specific humidity and pressure, then converts
+    #: back to specific humidity.
+    #: Only used by models calling to :func:`interpolate_met`.
+    interpolation_q_method: str = "linear-q"
```
mlshapiro, line 86: Does it complicate things if this is set by default to None? 

The default interpolation method should be whatever the overarching `interpolation_method` is -- the user could specify `bilinear` or `cubic` as the interpolation method for the whole grid as its written.

pycontrails/core/models.py
```diff
@@ -696,12 +723,13 @@ def interpolate_met(
     vector: GeoVectorDataset,
     met_key: str,
     vector_key: str | None = None,
+    q_method: str = "linear-q",
```
mlshapiro, line 726: You could remove this input if you put this value in `interp_kwargs`.  You could then `pop` the key out so that the interface does not break downstream.

pycontrails/models/cocip/cocip.py
```diff
@@ -538,10 +538,11 @@ def _process_flight(self, met: MetDataset) -> None:
         scale_humidity = humidity_scaling is not None and "specific_humidity" not in self.source
         verbose_outputs = self.params["verbose_outputs"]
 
-        interpolate_met(met, self.source, "air_temperature", **self.interp_kwargs)
-        interpolate_met(met, self.source, "specific_humidity", **self.interp_kwargs)
-        interpolate_met(met, self.source, "eastward_wind", "u_wind", **self.interp_kwargs)
-        interpolate_met(met, self.source, "northward_wind", "v_wind", **self.interp_kwargs)
+        interp_kwargs = {"q_method": self.params["interpolation_q_method"], **self.interp_kwargs}
```
mlshapiro, line 541: oh interesting! You do include it in `interp_kwargs` here

pycontrails/models/humidity_scaling.py
```diff
@@ -568,20 +569,27 @@ def scale(
...
-    path = pathlib.Path(__file__).parent / "quantiles" / "iagos_quantiles.npy"
-    # FIXME: Recompute to avoid the divide by 100.0 here
-    return np.load(path, allow_pickle=False) / 100.0
+def _load_quantiles() -> pd.DataFrame:
+    """Load precomputed ERA5 and IAGOS quantiles.
 
-
-@functools.cache
-def _load_era5_ensemble_quantiles() -> npt.NDArray[np.float64]:
-    path = pathlib.Path(__file__).parent / "quantiles" / "era5_ensemble_quantiles.npy"
```
mlshapiro, line 579: I think you should create a directory `models/humidity_scaling/` and then include this file in there along with a sub module `humidity_scaling`. 

You can add an `__init__` to expose the methods/classes you need in there.

I'd like to keep static data alongside the models

pycontrails/models/quantiles/era5-quantiles.pq
mlshapiro, line 1: see previous comment about moving files into a single models directory `models/humidity_scaling/`

pyproject.toml
```diff
@@ -168,7 +168,7 @@ exclude = ["docs*", "tests*"]
 [tool.setuptools.package-data]
 "pycontrails" = ["py.typed"]
 "pycontrails.models.emissions" = ["static/*.csv"]
-"pycontrails.models" = ["quantiles/*.npy"]
+"pycontrails.models" = ["quantiles/era5-quantiles.pq"]
```
mlshapiro, line 171: Remember to update paths here!

pycontrails/core/models.py
```diff
@@ -553,11 +570,21 @@ def set_source_met(
...
+
             # interpolate GeoVectorDataset
             if isinstance(self.source, GeoVectorDataset):
-                interpolate_met(self.met, self.source, met_key, **self.interp_kwargs)
+                interpolate_met(
+                    self.met, self.source, met_key, q_method=q_method, **self.interp_kwargs
+                )
                 continue
 
+            if q_method != "linear-q":
```
mlshapiro, line 582: this could be `if self.interp_kwargs["q_method"] is not None:`

pycontrails/core/models.py
```diff
@@ -731,25 +762,271 @@ def interpolate_met(
...
+        except KeyError as exc:
+            raise KeyError(f"No variable key '{met_key}' in 'met'.") from exc
+
+        out = vector.intersect_met(mda, **interp_kwargs)
+
+    vector[vector_key] = out
+    return out
+
+
+def _extract_q(met: MetDataset, met_key: str, q_method: str) -> tuple[MetDataArray, bool]:
```
mlshapiro, line 791: I think you could move everything below to the `core.interpolation` module to make it clear.

pycontrails/core/models.py
```diff
@@ -731,25 +762,271 @@ def interpolate_met(
...
+    vector_key = vector_key or met_key
 
     if (out := vector.get(vector_key, None)) is not None:
         return out
 
     if met is None:
-        raise ValueError(f"No variable key {vector_key} in ``vector`` and ``met`` is None")
+        raise ValueError(f"No variable key '{vector_key}' in 'vector' and 'met' is None")
+
+    if met_key in ("q", "specific_humidity"):
```
mlshapiro, line 773: i'd say `if met_key in ("q", "specific_humidity") and interp_kwargs["q_method"] is not None:`

pycontrails/core/models.py
mlshapiro: When we decide to set a default, then this would allow the user to turn it off by setting the value to None -- that feels more intuitive.

pycontrails/core/models.py
```diff
@@ -731,25 +762,271 @@ def interpolate_met(
...
+    Returns
+    -------
+    mda : MetDataArray
+        MetDataArray of specific humidity transformed for interpolation.
+    level : npt.NDArray[np.float_]
+        Transformed levels for interpolation.
+    """
+    da = mda.data
+
+    if q_method == "log-q-log-p":
```
mlshapiro, line 856: it feels like there should be a separate function for each of these, although I guess that starts to get verbose.

pycontrails/core/models.py
```diff
@@ -731,25 +762,271 @@ def interpolate_met(
...
 
     if (out := vector.get(vector_key, None)) is not None:
         return out
 
     if met is None:
-        raise ValueError(f"No variable key {vector_key} in ``vector`` and ``met`` is None")
+        raise ValueError(f"No variable key '{vector_key}' in 'vector' and 'met' is None")
+
+    if met_key in ("q", "specific_humidity"):
+        mda, log_applied = _extract_q(met, met_key, q_method)
```
mlshapiro, line 774: this feels like it should be within `interpolate_gridded_specific_humidity` and then you can get rid of `log_applied` by clearly separating the approaches into different sub-functions

pycontrails/core/models.py
```diff
@@ -731,25 +762,271 @@ def interpolate_met(
...
+        Whether or not a log transform was applied to specific humidity.
+    **interp_kwargs : Any,
+        Additional keyword only arguments passed to `intersect_met`.
+
+    Returns
+    -------
+    np.ndarray
+        Interpolated values.
+    """
+    if q_method == "linear-q":
```
mlshapiro, line 919: This may be clearer if you had it separated by method from the start. There may be some shared logic I'm not seeing, but I think you could have your `extract_q` and `prepare_q` separately here depending on method.

In particular I don't like the `log_applied` boolean getting passed, but I understand if you need the flag.

mlshapiro (approved):
All looks good. 

I added some preferences on how we specify this (i.e. I think this should be in `interp_kwargs`) but otherwise I don't have any problem caching the log or anything else!

pycontrails/core/models.py
zebengberg: Thanks -- I replaced "linear-q" with None.

c190074612edb81b7716e1c83bc6034f97652be9

pycontrails/core/models.py
zebengberg: Thanks!

fe10a1df11aed5cedc1afe767701ef1a869b7059

pycontrails/core/models.py
zebengberg: fe10a1df11aed5cedc1afe767701ef1a869b7059

pycontrails/core/models.py
zebengberg: fe10a1df11aed5cedc1afe767701ef1a869b7059

pycontrails/core/models.py
zebengberg: fe10a1df11aed5cedc1afe767701ef1a869b7059

pycontrails/models/cocip/cocip.py
zebengberg: Yeah ... you can see I thought about this previously, but something wasn't quite working. Anyway, this is a little cleaner now.

91a567cf0230cca62bf9e2527107008cbb69992a

pycontrails/core/models.py
zebengberg: c190074612edb81b7716e1c83bc6034f97652be9

pycontrails/models/humidity_scaling.py
zebengberg: 1c61242657a9ccab0245473d6feaa4816e5fd7a2

pyproject.toml
zebengberg: 1c61242657a9ccab0245473d6feaa4816e5fd7a2

pycontrails/models/quantiles/era5-quantiles.pq
zebengberg: 1c61242657a9ccab0245473d6feaa4816e5fd7a2

pycontrails/core/models.py
zebengberg: bc6ff61fc94886d2ec84c6af329e1ebd21bcb31d

pycontrails/core/models.py
zebengberg: I've thought about this ... but it's not as easy as it sounds. We're calling `interpolate_gridded_specific_humidity` in a two different places, and the `log_applied` flag is determine upstream of this in one case. I agree that this is a little ugly as written, but I'm going to keep as is for now.

pycontrails/core/models.py
zebengberg: See comment above ... I'm going to keep the log_applied flag for now.

PR #68, opened 2023-06-29
Bugfix/v0.44

zebengberg:
### Breaking changes

- By default, call `xr.open_mfdataset` with `lock=False` in the `MetDataSource.open_dataset` method. This helps alleviate a `dask` threading issue similar to [this GitHub issue](https://github.com/pydata/xarray/issues/4406).

### Fixes

- Support `MetDataset` source in the `HistogramMatching` humidity scaling model. Previously only `GeoVectorDataset` sources were explicitly supported.

## Tests

- [x] CI tests pass


mlshapiro (approved):
@zebengberg I took a quick review here, all looks fine to me. Helpful to see we prefer `np.clip`, and I will have to ask you about where we should use `dask` rather than `np` in the future.

PR #75, opened 2023-08-09
Feature/dry advection

zebengberg:
Support "dry advection" simulation.

### Features

- Add new `DryAdvection` model to simulate sediment-free advection of an aircraft's exhaust plume. This model is experimental and may change in future releases. By default, the current implementation simulates plume geometry as a cylinder with an elliptical cross section (the same geometry assumed in CoCiP). Wind shear perturbs the ellipse azimuth, width, and depth over the plume evolution. The `DryAdvection` model may also be used to simulate advection without wind-shear effects by setting the model parameters `azimuth`, `width`, and `depth` to None.
- Add optional `fill_value` parameter to `VectorDataset.sum`.

### Internals

- Update handling of `GeoVectorDataset.required_keys` the `GeoVectorDataset` constructor. Add class variable `GeoVectorDataset.vertical_keys` for handing the vertical dimension.
- Rename `CocipParam.max_contrail_depth` -> `CocipParam.max_depth`.
- Add `units.dt_to_float_seconds` function to convert `np.timedelta64` to `float` seconds.

## Tests

- [x] CI tests pass

## Reviewer

@mlshapiro 
@ktait98

### DIFF
--- a/pycontrails/core/vector.py
+++ b/pycontrails/core/vector.py
@@ -620,8 +620,8 @@ class VectorDataset:
         infer_attrs : bool, optional
             If True, infer attributes from the first VectorDataset in the list.
         fill_value : float, optional
-            Fill value to use when concatenating arrays, by default None, which raises an error
-            if incompatible keys are found.
+            Fill value to use when concatenating arrays. By default None, which raises
+            an error if incompatible keys are found.
 
         Returns
         -------
--- a/pycontrails/core/vector.py
+++ b/pycontrails/core/vector.py
@@ -609,6 +609,7 @@ class VectorDataset:
         cls: Type[VectorDatasetType],
         vectors: Sequence[VectorDataset],
         infer_attrs: bool = True,
+        fill_value: float | None = None,
     ) -> VectorDatasetType:
         """Sum a list of :class:`VectorDataset` instances.
 
@@ -618,6 +619,9 @@ class VectorDataset:
             List of :class:`VectorDataset` instances to concatenate.
         infer_attrs : bool, optional
             If True, infer attributes from the first VectorDataset in the list.
+        fill_value : float, optional
+            Fill value to use when concatenating arrays, by default None, which raises an error
+            if incompatible keys are found.
 
         Returns
         -------
@@ -651,13 +655,24 @@ class VectorDataset:
         """
         if not vectors:
             return cls()
-        keys = vectors[0].data.keys()
-        for v in vectors[1:]:
-            if v.data.keys() != keys:
-                raise KeyError("Summands have incompatible keys.")
+
+        keys: Iterable[str]
+        if fill_value is None:
+            keys = vectors[0].data.keys()
+            for v in vectors[1:]:
+                if v.data.keys() != keys:
+                    raise KeyError("Summands have incompatible keys.")
+        else:
+            keys = set().union(*[v.data.keys() for v in vectors])
+
+        def _get(k: str, v: VectorDataset) -> np.ndarray:
+            try:
+                return v[k]
+            except KeyError:
+                return np.full(v.size, fill_value)
 
         def concat(key: str) -> np.ndarray:
-            values = [v[key] for v in vectors]
+            values = [_get(key, v) for v in vectors]
             return np.concatenate(values)
 
         data = {key: concat(key) for key in keys}
--- a/pycontrails/core/vector.py
+++ b/pycontrails/core/vector.py
@@ -668,6 +668,8 @@ class VectorDataset:
             keys = set().union(*[v.data.keys() for v in vectors])
 
         def _get(k: str, v: VectorDataset) -> np.ndarray:
+            # Could also use VectorDataset.get() here, but we want to avoid creating
+            # an unused array if the key is present in the VectorDataset.
             try:
                 return v[k]
             except KeyError:
--- a/pycontrails/core/vector.py
+++ b/pycontrails/core/vector.py
@@ -225,7 +225,7 @@ def _empty_vector_dict(keys: Iterable[str]) -> VectorDataDict:
     VectorDataDict
         Empty :class:`VectorDataDict` instance.
     """
-    keys = keys or set()
+    keys = keys or ()
     data = VectorDataDict({key: np.array([]) for key in keys})
 
     # The default dtype is float64
@@ -1096,7 +1096,7 @@ class GeoVectorDataset(VectorDataset):
     time : npt.ArrayLike, optional
         Time data.
         Expects an array of DatetimeLike values,
-        or array that can proccessed with :meth:`pd.to_datetime`.
+        or array that can processed with :meth:`pd.to_datetime`.
         Defaults to None.
     attrs : dict[Hashable, Any] | AttrDict, optional
         Additional properties as a dictionary.
@@ -1115,11 +1115,10 @@ class GeoVectorDataset(VectorDataset):
     """
 
     #: Required keys for creating GeoVectorDataset
-    required_keys: set[str] = {
-        "latitude",
-        "longitude",
-        "time",
-    }
+    required_keys = "longitude", "latitude", "time"
+
+    #: At least one of these vertical-coordinate keys must also be included
+    vertical_keys = "altitude", "level", "altitude_ft"
 
     def __init__(
         self,
@@ -1146,12 +1145,12 @@ class GeoVectorDataset(VectorDataset):
             and level is None
             and time is None
         ):
-            keys = self.required_keys.union(["altitude"])
+            keys = *self.required_keys, "altitude"
             data = _empty_vector_dict(keys)
 
         super().__init__(data=data, attrs=attrs, copy=copy, **attrs_kwargs)
 
-        # using the self[key] syntax specifically to run qc on assigment
+        # using the self[key] syntax specifically to run qc on assignment
         if longitude is not None:
             self["longitude"] = np.array(longitude, copy=copy)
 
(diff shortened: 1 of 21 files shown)


pycontrails/core/vector.py
```diff
@@ -225,7 +225,7 @@ def _empty_vector_dict(keys: Iterable[str]) -> VectorDataDict:
     VectorDataDict
         Empty :class:`VectorDataDict` instance.
     """
-    keys = keys or set()
+    keys = keys or ()
```
mlshapiro, line 228: good catch

pycontrails/core/vector.py
```diff
@@ -651,13 +655,26 @@ def sum(
...
+            keys = vectors[0].data.keys()
+            for v in vectors[1:]:
+                if v.data.keys() != keys:
+                    diff = set(v.data.keys()).symmetric_difference(keys)
+                    raise KeyError(f"Summands have incompatible keys. Difference: {diff}")
+
+        else:
+            keys = set().union(*[v.data.keys() for v in vectors])
+
+        def _get(k: str, v: VectorDataset) -> np.ndarray:
```
mlshapiro, line 670: do we not have a VectorDataset.get() that would do this?

pycontrails/models/cocip/cocip.py
```diff
@@ -712,8 +712,9 @@ def _simulate_wake_vortex_downwash(self, met: MetDataset) -> None:
         wingspan = self._sac_flight.get_data_or_attr("wingspan")
 
         # get the pressure level `dz_m` lower than element pressure
-        air_pressure_lower = thermo.p_dz(air_temperature, air_pressure, self.params["dz_m"])
-        level_lower = air_pressure_lower / 100
+        dz_m = self.params["dz_m"]
+        air_pressure_lower = thermo.p_dz(air_temperature, air_pressure, dz_m)
```
mlshapiro, line 716: we should really rename `thermo.p_dz` to something more expressive, even `pressure_dz` would be better.

pycontrails/models/cocip/cocip_params.py
```diff
@@ -61,11 +61,12 @@ class CocipParams(ModelParams):
...
-    #: Integrate initial_contrails with time steps of dt
+    #: Apply Euler's method with a fixed step size of ``dt_integration``. Advected waypoints
+    #: are interpolated against met data once each ``dt_integration``.
     dt_integration: np.timedelta64 = np.timedelta64(30, "m")
 
-    #: Difference in altitude between top and bottom layer for stratification calculations (m)
-    #: Used to approximate derivative of "lagrangian_tendency_of_air_pressure" layer
+    #: Difference in altitude between top and bottom layer for stratification calculations,
+    #: [:math:`m`]. Used to approximate derivative of "lagrangian_tendency_of_air_pressure" layer.
     dz_m: float = 200.0
```
mlshapiro, line 70: I wish this parameter was named more clearly too, but I don't need to over do it. 

Does the 2012 paper use a specific variable? At least we could make sure we're consistent with the paper.

pycontrails/models/cocip/contrail_properties.py
```diff
@@ -1213,22 +1214,22 @@ def plume_temporal_evolution(
     """
     # Convert dt to seconds value and use dtype of other variables
     dtype = np.result_type(width_t1, depth_t1, sigma_yz_t1, dsn_dz_t1, diffuse_h_t1, diffuse_v_t1)
-    dt_s = dt / np.timedelta64(1, "s")
-    dt_s = dt_s.astype(dtype, copy=False)
+    dt_s = units.dt_to_float_seconds(dt, dtype)
```
mlshapiro, line 1217: Great! This seems like it should be `dt_to_seconds` since you are specifying the dtype here. I'll comment below in units.

mlshapiro: We should have added these utilities long ago.

pycontrails/physics/units.py
```diff
@@ -447,3 +447,26 @@ def lbs_to_kg(lbs: ArrayScalarLike) -> ArrayScalarLike:
         mass, kilograms [:math:`kg`]
     """
     return lbs * 0.45359
+
+
+def dt_to_float_seconds(
```
mlshapiro, line 452: I love this. I feel like this should be renamed `dt_to_seconds` with a default `dtype` of float64?

tests/unit/test_units.py
```diff
@@ -139,6 +139,7 @@ def test_mach_tas(rng: np.random.Generator):
             "m_to_longitude_distance",
             "tas_to_mach_number",
             "mach_number_to_tas",
+            "dt_to_float_seconds",
```
mlshapiro, line 142: flag here if change the name

pycontrails/models/dry_advection.py
```diff
@@ -0,0 +1,430 @@
...
+
+@dataclasses.dataclass
+class DryAdvectionParams(models.ModelParams):
+    """Parameters for the :class:`DryAdvection` model."""
+
+    #: Apply Euler's method with a fixed step size of ``dt_integration``. Advected waypoints
+    #: are interpolated against met data once each ``dt_integration``.
+    dt_integration: np.timedelta64 = np.timedelta64(30, "m")
+
+    #: Max age of contrail evolution.
```
mlshapiro, line 27: not "contrail"

pycontrails/models/dry_advection.py
```diff
@@ -0,0 +1,430 @@
...
+    max_age: np.timedelta64 = np.timedelta64(20, "h")
+
+    #: Difference in altitude between top and bottom layer for stratification calculations,
+    #: [:math:`m`]. Used to approximate derivative of "lagrangian_tendency_of_air_pressure" layer.
```
mlshapiro, line 31: We should add more here since "lagrangian tendency..." is opaque.  What about

> Used to approximate derivative of ``lagrangian_tendency_of_air_pressure`` (upward component of air velocity).

pycontrails/models/dry_advection.py
```diff
@@ -0,0 +1,430 @@
...
+    # If None, only pointwise advection is simulated without wind shear effects.
+    depth: float | None = 100.0
+
+    #: Initial plume direction, [:math:`m`]. Overridden by "azimuth" key on :attr:`source`.
+    # If None, only pointwise advection is simulated without wind shear effects.
+    azimuth: float | None = 0.0
+
+
+class DryAdvection(models.Model):
+    """Simulate "dry advection" of a plume with an elliptical cross section.
```
mlshapiro, line 51: "...of an **emissions** plume"

pycontrails/models/dry_advection.py
```diff
@@ -0,0 +1,430 @@
...
+    met : MetDataset
+        Meteorological data.
+    params : dict[str, Any]
+        Model parameters. See :class:`DryAdvectionParams` for details.
+    **kwargs : Any
+        Additional parameters passed as keyword arguments.
+    """
+
+    name = "dry_advection"
+    long_name = "Advection without sedimentation"
```
mlshapiro, line 80: "Emission plume advection without sedimentation"

mlshapiro (approved):
A few naming comments, but overall very clean.

The new model code looks very efficient, I'm glad we're starting to get some returns on all the infrastructure development of a year ago.

mlshapiro:
I'd love to get a benchmark from another group's dry advection code. Perhaps ask Scott/Kevin or @reyrem for some outputs they have?

pycontrails/core/vector.py
zebengberg: You're absolutely right (I forget about that here). One subtle but important difference is that when you call `VectorDataset.get("some_key", some_default)`, the `some_default` value is eagerly computed. In other words, when you call `VectorDataset.get("some_key", some_expensive_calculation())`, the expensive calculation takes place regardless of whether "some_key" is in the vector. In this case, we end up constructing a numpy array only to throw it away.

I think I'll keep the implementation as it stands, but include a comment.

mlshapiro: Should we update the other one to not be eager?

pycontrails/models/cocip/cocip.py
zebengberg: cbb961f1774b3081875f7e24f

pycontrails/core/vector.py
zebengberg: I don't think so -- the implementation gets too tricky. The lazy version would only be possible if the default argument was something like a callback that only gets evaluated if the key is not in the dict.

Something like this may work, but I'm still inclined not to include this in pycontrails. It's different from dict.get, and it may be over-engineered.

```python

class VectorDataset:
    ....

    def lazy_get(key, default):
        out = self.data.get(key, default)
        if out is default and callable(default):
            return default()
        return out
```

pycontrails/models/dry_advection.py
zebengberg: 5cc8de6ce58a54d6716e57b20715afa658af5848

tests/unit/test_units.py
zebengberg: 5f96a0a772b5603c69292a3f9b7d781ef9b3114d

pycontrails/physics/units.py
zebengberg: 5f96a0a772b5603c69292a3f9b7d781ef9b3114d

pycontrails/models/dry_advection.py
zebengberg: c69d26b1e2af7af0abaf07a9414ea71d8550007d

pycontrails/models/dry_advection.py
zebengberg: c69d26b1e2af7af0abaf07a9414ea71d8550007d

pycontrails/models/dry_advection.py
zebengberg: 47fdf46dfb902a208650e13d6a0ac6f5ebb23b25

Reyrem:
I could provide a benchmark for advected flight trace, but it seems that it is not exactly built the same way you designed the dry advection. 
I take into account sedimentation as a function of mean cristal size. Because we don't want to use any information about the ambiant humidity, it is not possible to use the standard way of computing the cristal size. I therefore used a regression method (fitted on cocip outputs) to estimate the cristal size as a function of time only. 
Feel free to ask me for anymore information !

mlshapiro:
> I could provide a benchmark for advected flight trace, but it seems that it is not exactly built the same way you designed the dry advection. I take into account sedimentation as a function of mean cristal size. Because we don't want to use any information about the ambiant humidity, it is not possible to use the standard way of computing the cristal size. I therefore used a regression method (fitted on cocip outputs) to estimate the cristal size as a function of time only. Feel free to ask me for anymore information !

This would be great. We thought about adding sedimentation. Would you be willing to include your model as parameter in this implementation?

Reyrem:
@mlshapiro I think so, but it might take some time... Is that ok for you?

mlshapiro:
Of course - we can just add to the board for now. See #82

PR #83, opened 2023-08-21
Feature/ps grid

zebengberg:
Implement a Poll-Schumann (`PSGrid`) theoretical aircraft performance over a grid.

### Breaking changes

- Move the `pycontrails.models.aircraft_performance` module to `pycontrails.core.aircraft_performance`.

### Features

- Implement the `PSGrid` model. For a given aircraft type and position, this model computes optimal aircraft performance at cruise conditions. In particular, this model can be used to estimate fuel flow, engine efficiency, and aircraft mass at cruise. In particular, the `PSGrid` model can now be used in conjunction with `CocipGrid` to simulate contrail formation over a grid.
- Refactor the `Emissions` model so that `Emissions.eval` runs with `source: GeoVectorDataset`. Previously, the `eval` method required a `Flight` instance for the `source` parameter. This change allows the `Emissions` model to run more seamlessly as a sub-model of a gridded model (ie, `CocipGrid`),
- No longer require `pycontrails-bada` to import or run the `CocipGrid` model. Instead, the `CocipGridParams.aircraft_performance` parameter can be set to any `AircraftPerformanceGrid` instance. This allows the `CocipGrid` model to run with any aircraft performance model that implements the `AircraftPerformanceGrid` interface.

### Internals

- Add optional `climb_descend_at_end` parameter to the `Flight.resample_and_fill` method. If True, the climb or descent will be placed at the end of each segment rather than the start.
- Define `AircraftPerformanceGridParams`, `AircraftPerformanceGrid`, and `AircraftPerformanceGridData` abstract interfaces for gridded aircraft performance models.
- Add `set_attr` parameter to `Models.get_source_param`.
- Better handle `source`, `source.attrs`, and `params` customizations in `CocipGrid`.
- Include additional classes and functions in the `pycontrails.models.emissions` module.
- Hardcode the paths to the static data files used in the `Emissions` and `PSModel` models. Previously these were configurable by model parameters.

## Tests

- [x] CI tests pass


trdean1 (approved):
Everything I went through in here looks great to me!  I left some minor comments that reflect the errors I came across when I tried to use it as a user initially.  I also played around with the grid model output a bit and think it looks pretty reasonable on first pass.

pycontrails/core/aircraft_performance.py
```diff
@@ -476,3 +490,62 @@ class AircraftPerformanceData:
...
+@dataclasses.dataclass
+class AircraftPerformanceGridParams(ModelParams):
+    """Parameters for :class:`AircraftPerformanceGrid`."""
+
+    #: Fuel type
+    fuel: fuel.Fuel = dataclasses.field(default_factory=fuel.JetA)
+
+    #: ICAO code designating simulated aircraft type.
+    #: Can be overridden by including ``aircraft_type`` attribute in source data
+    aircraft_type: str = "B737"
```
mlshapiro, line 509: At some point in the future, we should add an dictionary of all the ICAO codes. I'm often looking for this

pycontrails/core/aircraft_performance.py
```diff
@@ -476,3 +490,62 @@ class AircraftPerformanceData:
...
+        self, source: GeoVectorDataset | MetDataset | None = None, **params: Any
+    ) -> GeoVectorDataset | MetDataset:
+        """Evaluate the aircraft performance model."""
+
+
+@dataclasses.dataclass
+class AircraftPerformanceGridData(Generic[ArrayOrFloat]):
+    """Store the computed aircraft performance metrics for nominal cruise conditions."""
+
+    fuel_flow: ArrayOrFloat
```
mlshapiro, line 550: Should we add units here for sake of documentation?

pycontrails/core/fleet.py
```diff
@@ -134,18 +132,19 @@ def from_seq(
...
         for fl in seq:
-            cls._validate_fl(fl, seq[0].fuel, attrs, broadcast_numeric)
+            cls._validate_fl(fl, fuel, attrs, broadcast_numeric)
 
         # Surprisingly, it's faster to call np.concatenate on each variable
         # separately then it is to call pd.concat on the sequence of fl.dataframe.
         # We do this concatenation below.
         data = {var: np.concatenate([fl[var] for fl in seq]) for var in attrs["data_keys"]}
-        return cls(data=data, attrs=attrs, copy=False)
+        return cls(data=data, attrs=attrs, copy=False, fuel=fuel)
```
mlshapiro, line 147: I regret adding this as a class init param. Fuel should probably have been managed as an attr.

pycontrails/core/met.py
```diff
@@ -707,7 +707,7 @@ def get(self, key: str, default_value: Any = None) -> Any:
     def __setitem__(
         self,
         key: Hashable | list[Hashable] | Mapping,
-        value: MetDataArray | xr.DataArray | np.ndarray | xr.Variable,
+        value: Any,
```
mlshapiro, line 710: Why the generalization of type here?

pycontrails/models/cocip/cocip.py
```diff
@@ -596,17 +595,20 @@ def _process_emissions(self) -> None:
...
 
-        # currently only one emissions model
-        emissions = Emissions(copy_source=False)
+        # Call aircraft performance and Emissions models as needed
+        # NOTE: None of these sub-models actually do any met interpolation -- self.source already
+        # has all of the required met variables attached. Therefore, we don't need to worry about
+        # being consistent with passing in Cocip's interp_kwargs and humidity_scaling into
+        # the sub-models.
+        emissions = Emissions()
+        ap_model = self.params["aircraft_performance"]
```
mlshapiro, line 604: Would it be too verbose to call this `performance_model` or `aircraft_performance` just to make it clear throughout?

pycontrails/models/cocipgrid/cocip_grid_params.py
```diff
@@ -6,22 +6,13 @@
 
 import numpy as np
 
+from pycontrails.core.aircraft_performance import AircraftPerformanceGrid
 from pycontrails.core.fuel import Fuel, JetA
-
-# from pycontrails.models.aircraft_performance import AircraftPerformanceGrid
 from pycontrails.models.cocip.cocip_params import CocipParams
 
-try:
```
mlshapiro, line 14: This is awesome! Removing one more dependency

pycontrails/models/cocip/cocip.py
mlshapiro: I see this is consistent with `cocip_grid` - Fine leaving it as it

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -1543,16 +1582,13 @@ def calc_evolve_one_step(
...
-    The BADA database to use (BADA3 or BADA4) is determined by the ``bada_priority``
-    parameter.
+    """Calculate aircraft performance (AP) and emissions data.
 
     This function mutates the ``vector`` parameter in-place by setting keys:
-        - "true_airspeed": nominal value from BADA database to use for TAS
-        - "engine_efficiency": computed in :meth:`BADA.simulate_fuel_and_performance`
-        - "fuel_flow": compute in :meth:`BADA.simulate_fuel_and_performance`
-        - "nvpm_ei_n": computed in Emissions methods
+        - "true_airspeed": computed by the AP model
```
mlshapiro, line 1588: I would write out here "aircraft performance" model

mlshapiro (approved):
I did a first review without testing code, I'll give a test now!

pycontrails/models/ps_model/ps_grid.py
```diff
@@ -0,0 +1,469 @@
...
+        if self.get_source_param("aircraft_mass", set_attr=False) is not None:
+            raise NotImplementedError("The 'aircraft_mass' parameter must be None.")
+        if isinstance(self.source, Flight):
+            raise TypeError("PSGrid doesn't support Flight objects as source data. Use PSModel.")
+
+        # Extract the relevant source data
+        try:
+            aircraft_type = self.source.attrs["aircraft_type"]
+        except KeyError:
+            aircraft_type = self.params["aircraft_type"]
```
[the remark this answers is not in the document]
zebengberg: Ok! It might be a little better now. 98a4d0da3ebaef431059501970665c58dc2c1815

pycontrails/models/ps_model/ps_grid.py
```diff
@@ -0,0 +1,469 @@
...
+            - fuel_flow
+            - engine_efficiency
+
+        See Also
+        --------
+        :func:`ps_nominal_grid`
+        """
+        self.update_params(**params)
+        self.set_source(source)
+        self.require_source_type((GeoVectorDataset, MetDataset))
```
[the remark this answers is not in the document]
zebengberg: You're right ... what I had violates liskov substitution.

4d42420c2decb953a803cfa3e3e56e0c6a9ffbf4

pycontrails/models/ps_model/ps_grid.py
```diff
@@ -0,0 +1,469 @@
...
+        self.set_source_met()
```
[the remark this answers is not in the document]
zebengberg: Yeah ... I briefly considered this. Anytime we have a model which requires air temperature but a `met` parameter is not given, we could fallback on the ISA temperature. I think this is something the user should do manually ... perhaps a convenience function could help.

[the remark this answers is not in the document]
zebengberg: ef0ff64aadc9866f9f7991127f2d568ffbed51b6

pycontrails/core/met.py
zebengberg: For all instances of `value` except for `MetDataArray`, we simply call `xr.Dataset.__setitem__`. The signature there is `value: Any`, and I was noticing that we were being too strict here. For example, in the `ACCF` model, we had a line of code:

```
self.source[key] = (maCCFs.dim_order, arr.data)  # type: ignore
```

We don't allow tuple as a possible type of `value`, but `xr.Dataset.__setitem__` allows this.

pycontrails/models/cocip/cocip.py
zebengberg: 👍

pycontrails/core/fleet.py
zebengberg: I agree, although the `fuel` attribute isn't too much of a headache. We definitely could start using `attrs` as the source of truth for `fuel` and add a `DeprecationWarning` when `Flight.fuel` is accessed. Perhaps we could do this as part of a larger cleanup when we get some feedback from the community.

pycontrails/core/aircraft_performance.py
zebengberg: ccd293abe4bb4d50c063843b46e3b41eebebe273

PR #87, opened 2023-08-30
Bug fix/conditionally compute tau cirrus after met downselection

tsankar:
## Changes

- Adds "compute_tau_cirrus_in_model_init" parameter which determines when tau_cirrus is computed.
- In "auto" mode, computes tau_cirrus in init iff met is dask-backed. Otherwise, computes it after downselection.

#### Breaking changes
None

#### Fixes
- Memory explosion when using CoCiP on non-dask-backed weather data.
## Tests

-- GCP tests in test_cache.py fail locally, presumably because of auth issues?
- [ ] CI tests pass
-- Let me know the best way to run these if necessary!

## Reviewer

> @zebengberg

### DIFF
--- a/pycontrails/core/models.py
+++ b/pycontrails/core/models.py
@@ -92,6 +92,13 @@ class ModelParams:
     #: Call :meth:`_verify_met` on model instantiation
     verify_met: bool = True
 
+    #: Compute ``"tau_cirrus"`` variable in pressure-level met data during model
+    #: initialization. Must be one of ``"auto"``, ``True``, or ``False``. If set to
+    #: ``"auto"``, ``"tau_cirrus"`` will be computed during model initialization
+    #: iff the met data is dask-backed. Otherwise, it will be computed during model
+    #: evaluation after the met data is downselected.
+    compute_tau_cirrus_in_model_init: bool | str = "auto"
+
     #: Downselect input MetDataset to region around ``source``, if input.
     downselect_met: bool = True
 
--- a/pycontrails/core/models.py
+++ b/pycontrails/core/models.py
@@ -92,13 +92,6 @@ class ModelParams:
     #: Call :meth:`_verify_met` on model instantiation
     verify_met: bool = True
 
-    #: Compute ``"tau_cirrus"`` variable in pressure-level met data during model
-    #: initialization. Must be one of ``"auto"``, ``True``, or ``False``. If set to
-    #: ``"auto"``, ``"tau_cirrus"`` will be computed during model initialization
-    #: iff the met data is dask-backed. Otherwise, it will be computed during model
-    #: evaluation after the met data is downselected.
-    compute_tau_cirrus_in_model_init: bool | str = "auto"
-
     #: Downselect input MetDataset to region around ``source``, if input.
     downselect_met: bool = True
 
--- a/pycontrails/models/cocip/cocip.py
+++ b/pycontrails/models/cocip/cocip.py
@@ -279,7 +279,8 @@ class Cocip(Model):
         super().__init__(met, params=params, **params_kwargs)
 
         shift_radiation_time = self.params["shift_radiation_time"]
-        met, rad = process_met_datasets(met, rad, shift_radiation_time)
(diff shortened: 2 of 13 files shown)


pycontrails/core/models.py
```diff
@@ -92,6 +92,13 @@ class ModelParams:
     #: Call :meth:`_verify_met` on model instantiation
     verify_met: bool = True
 
+    #: Compute ``"tau_cirrus"`` variable in pressure-level met data during model
+    #: initialization. Must be one of ``"auto"``, ``True``, or ``False``. If set to
+    #: ``"auto"``, ``"tau_cirrus"`` will be computed during model initialization
+    #: iff the met data is dask-backed. Otherwise, it will be computed during model
+    #: evaluation after the met data is downselected.
+    compute_tau_cirrus_in_model_init: bool | str = "auto"
```
zebengberg, line 100: It may be better to move this to `CocipParams` since its only used in `Cocip` and `CocipGrid`. I think `CocipGridParams` subclasses `CocipParams`, so you wouldn't need to duplicate there.

pycontrails/models/cocip/cocip.py
```diff
@@ -1274,6 +1279,7 @@ def _fill_empty_flight_results(self, return_list_flight: bool) -> Flight | list[
 def process_met_datasets(
     met: MetDataset,
     rad: MetDataset,
+    compute_tau_cirrus: bool | str = "auto",
     shift_radiation_time: np.timedelta64 | None = None,
```
zebengberg, line 1283: Does it make sense to remove the default values for `compute_tau_cirrus` (and for `shift_radiation_time`)? If we only ever pass model parameters into this function, then we may be less likely to create future bugs if these parameters *don't* take defaults. But if we're calling `process_met_datasets` in places where we don't specify `compute_tau_cirrus` or `shift_radiation_time`, leave as is.

zebengberg (approved):
Thank you Tharun, this is great!

I made some a commit straight to main while your PR was opened ... can you rebase off main then re-push?

pycontrails/models/cocip/cocip.py
```diff
@@ -1310,9 +1319,20 @@ def process_met_datasets(
...
+            f" True, False, or 'auto'. Found {compute_tau_cirrus}."
+        )
     if "tau_cirrus" not in met.data:
         met.ensure_vars(Cocip.met_variables)
-        met.data["tau_cirrus"] = tau_cirrus.tau_cirrus(met)
+        if (
+            compute_tau_cirrus == "auto"
+            and met.data["specific_humidity"].chunks is not None
+            or isinstance(compute_tau_cirrus, bool)
+            and compute_tau_cirrus
```
zebengberg, line 1333: I had to think about this for a minute ... I thought there could be someway to simplify this clause before coming around to the way you have it.

pycontrails/core/models.py
tsankar: Thanks, didn't realize that `CocipGridParams` was a subclass of `CocipParams`! Moved it to `CocipParams`, this way makes more sense.

pycontrails/models/cocip/cocip.py
tsankar: The only other place we call `process_met_datasets` is in [tests/benchmark/north-atlantic-study/validate.py](https://github.com/contrailcirrus/pycontrails/blob/9e062cff5ed5f1dc0956e56b6a3ac21958d41405/tests/benchmark/north-atlantic-study/validate.py#L148), is this benchmark test still used? I agree that it would be safer to not include defaults here.

pycontrails/models/cocip/cocip.py
tsankar: Yeah, I spent embarrassingly long thinking about this - sometimes Python implicit bool casting is not your friend :)

pycontrails/models/cocip/cocip.py
zebengberg: We've been meaning to run the benchmark tests occasionally in the github CI. That hasn't happened yet, and so I think there are many slightly broken things in the benchmark directory at this point. No need to worry about that!

PR #93, opened 2023-09-13
Feature/step climb

trdean1:
## Changes

> Changes `_altitude_interpolation` method that is called within `resample_and_fill`
> When interpolating altitude, all descents go at end of segments. 
> Climbs go at the beginning of segments, unless the gap is greater than two hours in length, then it goes in the middle

#### Breaking changes

May cause some decreases in fuel consumption in longer flights with large gaps in ADSB coverage (e..g transoceanic)

#### Internals

> Keeps the flag `climb_descend_at_end` in `resample_and_fill` that always places altitude change at end of segment
> Added test case to cover a flight with long and short segments that both climb and descend
> Changed the test case `test_geodesic_interpolation` so that test runs on a flight that does not change altitude

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg 
> @roger-teoh


pycontrails/core/flight.py
```diff
@@ -1453,38 +1455,87 @@ def _altitude_interpolation(
     end_na_idxs = np.flatnonzero(end_na)
     na_group_size = end_na_idxs - start_na_idxs
 
-    # Form array of cumulative altitude values if the flight were to climb
-    # at nominal_rocd over each group of nan
+    s = pd.Series(altitude)
 
     if climb_or_descend_at_end:
```
zebengberg, line 1460: You may want to create two helper functions depending on `climb_or_descend_at_end`. So this may look like:

```python
if climb_or_descend_at_end:
   return _altitude_interpolation_climb_descend_end(altitude, s, na_group_size, other_params_needed, ...)
return _altitude_interpolation_climb_descend_middle(altitude, s, na_group_size, other_params_needed, ...)
```

pycontrails/core/flight.py
```diff
@@ -1453,38 +1455,87 @@ def _altitude_interpolation(
...
+            # Skip short segments and segments that do not climb
+            if not step_group:
+                continue
+            if altitude[start_na_idxs[i]] >= altitude[end_na_idxs[i]]:
+                continue
+
+            # We have a long climbing segment.  Keep first half of segment at the starting
+            # altitude, then climb at mid point. Adjust indicies computed before accordingly
+            is_odd = na_group_size[i] % 2
+            na_group_size[i] = math.floor(na_group_size[i] / 2)
```
zebengberg, line 1504: One way to avoid importing math:

```python
quot, is_odd = divmod(na_group_size[i], 2)
na_group_size[i] = quot
...
```

pycontrails/core/flight.py
```diff
@@ -1453,38 +1455,87 @@ def _altitude_interpolation(
...
+            start_na_idxs[i] += nan_fill_size
+
+    # Use pandas to forward and backfill altitude values
+    s_ff = s.ffill()
+    s_bf = s.bfill()
+
+    # Form array of cumulative altitude values if the flight were to climb
+    # at nominal_rocd over each group of nan
+    cumalt_list = []
+    for i, size in enumerate(na_group_size):
```
zebengberg, line 1517: You may want to use `zip` instead of `enumerate` here:

```python
for start_na_idx, end_na_idx, size in zip(start_na_idxs, end_na_idxs, na_group_size):
   ...
```

pycontrails/core/flight.py
```diff
@@ -1453,38 +1455,87 @@ def _altitude_interpolation(
...
+        if altitude[start_na_idxs[i]] <= altitude[end_na_idxs[i]]:
+            cumalt_list.append(np.arange(1, size, dtype=float))
+        else:
+            cumalt_list.append(np.flip(np.arange(1, size, dtype=float)))
```
zebengberg, line 1521: You could avoid the `np.flip` (here and earlier as well) with: `np.arange(size, 1, -1, dtype=float)`

tests/unit/test_flight.py
```diff
@@ -468,6 +470,45 @@ def _check_rocd(_fl: Flight, nominal_rocd: float = constants.nominal_rocd) -> np
...
+        ),
+        altitude=np.array([0, 5000, 10000, 9000, 5000]),
+    )
+
+    fl1 = fl_alt.resample_and_fill()
+    assert _check_rocd(fl1)
+    # The first segment's climb should be at the start
+    assert fl1["altitude"][1] > 0
+    # The second segment's climb should be in the middle
+    assert abs(fl1["altitude"][61] - 5000.0) < 1e-9
```
zebengberg, line 504: Another way to write this is with:

```python
assert fl1["altitude"][61] == pytest.approx(5000.0, abs=1e-9)
```

What you have is also really explicit!

zebengberg (approved):
Great job! So many tricky details here.

All of my feedback is completely optional. The only thing I feel slightly stronger about is splitting `_altitude_interpolation` into two (or more) helper functions ... the current implementation is pretty huge.

pycontrails/core/flight.py
trdean1: It would actually be `np.arange(size-1, 0, -1, dtype=float)` - I made that same mistake the first time and prefer the `flip` for readability.

PR #96, opened 2023-09-29
Critical bug fix on `correct_fuel_flow` of PS model

roger-teoh:
Closes #XX

## Changes

Resolved bug in the guard rails of the PS model, where the maximum fuel flow at certain conditions could be greater than the maximum fuel flow that can be supplied by the engine. 

#### Fixes
- Fix `equivalent_fuel_flow_rate_at_cruise` function in pycontrails.physics.jet
- Strengthen `correct_fuel_flow` in PS model to account for descent conditions, and compensate for the lack of PTF files for the PS model. 

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/tests/unit/test_ps_model.py
+++ b/tests/unit/test_ps_model.py
@@ -8,7 +8,7 @@ import pytest
 import xarray as xr
 
 import pycontrails.models.ps_model.ps_model as ps
-from pycontrails.core import Flight, GeoVectorDataset, MetDataset
+from pycontrails import Flight, FlightPhase, GeoVectorDataset, MetDataset
 from pycontrails.models.ps_model import PSGrid, ps_nominal_grid
 from pycontrails.physics import jet, units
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_ps_model.py
```diff
@@ -130,6 +130,7 @@ def test_ps_model() -> None:
         mach_number,
         atyp_param.ff_idle_sls,
         atyp_param.ff_max_sls,
+        FlightPhase.CRUISE,
```
roger-teoh, line 133: @zebengberg Should this be just `FlightPhase` instead of `FlightPhase.CRUISE`? Because ps.correct_fuel_flow looks at FlightPhase.descent?

zebengberg: We're okay here. You added a new parameter to `correct_fuel_flow`, so we need to put something here when we call the function in our tests. This parameter needs to be an instance of `FlightPhase` or an array of `FlightPhase` instances, not the type. Setting it to anything value besides `FlightPhase.descent` keeps the previous behavior, which is what we want to keep this test checking what it was designed for.

Issue #64, opened 2023-06-15
Wrong OLR weather value in ACCF contrail


Reyrem:
**Description**

The contrail accf by day is not using the right weather variable for outgoing longwave radiation. Climaccf is using "top net thermal radiation (ttr)" in J m^-2 from ERA5 single level dataset. It is then divided in their implementation by the time interval (3600s in this case) to convert it to the mean flux on this time interval (W m^-2). Climaccf should therefore be given the value in J m^-2, and not the one in W m^-2.

This is the expected result with Climaccf : 
![climaccf_contrailACCF](https://github.com/contrailcirrus/pycontrails/assets/73607302/dce59ae7-9402-48f1-9dc3-138b6681e84b)

This is the result with Pycontrail accf:
![pycontrail_contrailACCF](https://github.com/contrailcirrus/pycontrails/assets/73607302/a5774564-f986-4ab9-9e49-6dabd3f8c33d)

This is the result with climaccf with the wrong OLR values :
![climaccf_contrailACCF_wrong_OLR](https://github.com/contrailcirrus/pycontrails/assets/73607302/5a7b85ec-6b9c-4cf6-adf4-4fb95613d1c4)

We think this is the main difference between the two results.

**Details**

- Version: 		0.42.0
- Module:		models.accf
- OS: 			windows

**Steps to Reproduce**

1. Run the accfs with pycontrail
2. Compare with Climaccf with both the weather variable in J/m-² and in W/m-²

mlshapiro:
This is super helpful, thanks @Reyrem . 

We made the choice to modify the accumulated values on load because it made downstream modeling simpler, but only for the internally built models. I think it was an incorrect choice, so we'll find a way to stop modifying the input meteorology by default.

trdean1:
I just ran across this issue in some comparison work I was doing.  A work around to this would be to multiply by `3600*step_size` after [this line](https://github.com/contrailcirrus/pycontrails/blob/2c1b92fd45e3e24d6b6676c9e0382a9d2df8e483/pycontrails/models/accf.py#L309) in our wrapper code.  

@mlshapiro is there any objection to doing this now and removing it once we change how the met data gets loaded?

zebengberg:
Fixed in #104

PR #104, opened 2023-10-19
Feature/met clarity

zebengberg:
### Breaking changes

- When running `Cocip` and other `pycontrails` models, the `met` and `rad` parameter must now contain predefined metadata attributes `provider`, `dataset`, and `product` describing the met source. An error will now be raised in `Cocip` if these attributes are not present.
- Deprecate passing arbitrary `kwargs` into the `MetDataArray` constructor.
- No longer convert accumulated radiation data to average instantaneous data in `ERA5` and `HRES` interfaces. This logic is now handled downstream by the model (e.g., `Cocip`). This change allows for more flexibility in the `rad` data passed into the model and avoids unnecessary computation in the `MetDataSource` interfaces.
- Add new `MetDataSource.set_met_source_metadata` abstract method. This should be called within the implementing class `open_metdataset` method.
- No longer take a finite difference in the time dimension for HRES radiation data. This is now also handled natively in `Cocip`.
- No longer convert relative humidity from a percentage to a fraction in `ERA5` and `HRES` interfaces.
- Require the `HRES` `stream` parameter to be one of `["oper", "enfo"]`. Require the `field_type` parameter to be one of `["fc", "pf", "cf", "an"]`.
- Remove the `steps` and `step_offset` properties in the `GFSForecast` interface. Now the `timesteps` attribute is the only source of truth for determining AWS S3 keys. Change the `filename` method to take in a `datatime` timestep instead of an `int` step. No longer assign the first step radiation data to the zeroth step.
- Change the return type of `ISSR.eval`, `SAC.eval`, and `PCR.eval` from `MetDataArray` to `MetDataset`. This is more consistent with the return type of other `pycontrails` models and more closely mirrors the behavior of vector models. Set output `attrs` metadata on the global `MetDataset` instead of the individual `MetDataArray` in each case.

### Features

- Rewrite parts of the `pycontrails.core.datalib` module for higher performance and readability.
- Add optional `attrs` and `attrs_kwargs` parameters to `MetDataset` constructor. This allows the user to customize the attributes on the underlying `xarray.Dataset` object. This update makes `MetDataset` more consistent with `VectorDataset`.
- Add three new properties `provider_attr`, `dataset_attr`, and `product_attr` to `MetDataset`. These properties give metadata describing the underlying meterological data source.
- Add new `Model.transfer_met_source_attrs` method for more consistent handling of met source metadata on the `source` parameter passed into `Model.eval`.
- No longer require `geopotential` data when computing `tau_cirrus`. If neither `geopotential` nor `geopotential_height` are available, geopotential is approximated from the geometric height. No longer require geopotential on the `met` parameter in `Cocip` or `CocipGrid`.
- Remove the `Cocip` `shift_radiation_time` parameter. This is now inferred directly from the `rad` metadata. An error is raised if the necessary metadata is not present.
- Allow `Cocip` to run with both instantaneous (`W m-2`) and accumulated (`J m-2`) radiation data.
- Allow `Cocip` to run with accumulated ECMWF HRES radiation data.

### Fixes

- Correct radiation unit in the `ACCF` wrapper model [#64]. Both instantaneous (`W m-2`) and accumulated (`J m-2`) radiation data are now supported, and the `ACCF` wrapper will handle each appropriately.
- Avoid unnecessary writing and reading of temporary files in `ERA5.cache_dataset` and `HRES.cache_dataset`.
- Fix timestep resolution bug in `GFSForecast`. When the `grid` parameter is 0.5 or 1.0, forecasts are only available every 3 hours. Previously, the `timesteps` property would define an hourly timestep.

### Internals

- Include `name` parameter in `MetDataArray` constructor.
- Make the `coordinates.slice_domain` function slightly more performant by explicitly dropping nan values from the `request` parameter.
- Round unwieldy floating point numbers in `GeoVectorDataset._display_attrs`.
- Remove the `ecmwflibs` package from the `ecmwf` optional dependencies.
- Add NPY to `ruff` rules.


pycontrails/core/datalib.py
```diff
@@ -505,6 +526,19 @@ def open_metdataset(
...
+    @abc.abstractmethod
+    def set_met_source_metadata(self, ds: xr.Dataset | MetDataset) -> None:
+        """Set met source metadata on ``ds.attrs``.
+
+        This is called within the :meth:`open_metdataset` method to set metadata
+        on the returned :class:`MetDataset` instance.
+
+        Parameters
+        ----------
+        met : xr.Dataset | MetDataset
```
mlshapiro, line 538: `ds` is the parameter here, not `met`

pycontrails/core/datalib.py
```diff
@@ -505,6 +526,19 @@ def open_metdataset(
         :func:`xarray.open_mfdataset`
         """
 
+    @abc.abstractmethod
+    def set_met_source_metadata(self, ds: xr.Dataset | MetDataset) -> None:
```
mlshapiro, line 530: Why not call this `set_attrs` or `set_metadata`? A little less verbose

pycontrails/core/met.py
```diff
@@ -593,6 +593,16 @@ class MetDataset(MetBase):
         interval ``[-180, 180]``. Defaults to False.
     copy : bool, optional
         Copy data on construction. Defaults to True.
+    attrs : dict[str, Any], optional
```
mlshapiro, line 596: Fantastic! Thank you for cleaning this up

pycontrails/core/met.py
```diff
@@ -593,6 +593,16 @@ class MetDataset(MetBase):
...
+        Attributes to add to :attr:`data.attrs`. Defaults to None.
+        Generally, pycontrails :class:`Models` may use the following attributes:
+
+        - ``provider``: Name of the data provider (e.g. "ECMWF").
```
mlshapiro, line 600: I wonder if we want to prefix this attrs somehow? e.g. `pycontrails_provider`, `pycontrails_dataset` ... This would namespace the attrs and hopefully make it clear that the distinction is only meaningful in the context of our library?

pycontrails/core/met.py
```diff
@@ -1022,6 +1037,84 @@ def to_vector(self, transfer_attrs: bool = True) -> vector_module.GeoVectorDatas
...
 
+    def _get_attr_template(
+        self,
+        name: str,
+        supported: tuple[str, ...],
+        examples: dict[str, str],
+    ) -> str:
+        """Look up an attribute with a custom error message."""
+        try:
+            out = self.attrs[name].upper()
```
mlshapiro, line 1048: Why `upper()`?

Edit: OK i see, you are comparing to the pre-defined attrs. I would put an `.upper()` on supported below, so you re just making the comparison case insensitive.

pycontrails/core/met.py
```diff
@@ -1022,6 +1037,84 @@ def to_vector(self, transfer_attrs: bool = True) -> vector_module.GeoVectorDatas
...
+            msg = f"Specify '{name}' attribute on underlying dataset."
+
+            for i, (k, v) in enumerate(examples.items()):
+                if i == 0:
+                    msg = f"{msg} For example, set attrs['{name}'] = '{k}' for {v}."
+                else:
+                    msg = f"{msg} Set attrs['{name}'] = '{k}' for {v}."
+            raise KeyError(msg) from e
+
+        if out not in supported:
```
mlshapiro, line 1059: I would make `supported` all `.upper()` to make it truly case insensitive

pycontrails/core/met.py
mlshapiro: Maybe rename this `_get_pycontrails_attr_template` so that it specifies this isn't a generic `get_attr` thing?

pycontrails/core/met.py
```diff
@@ -1218,13 +1312,18 @@ def __init__(
...
         validate: bool = True,
+        name: Hashable | None = None,
         **kwargs: Any,
     ) -> None:
         # init cache
         self.cachestore = cachestore
 
         # try to create DataArray out of input data and **kwargs
         if not isinstance(data, xr.DataArray):
+            DeprecationWarning(
```
mlshapiro, line 1323: How should we tag these for revision in the future?

pycontrails/core/met.py
```diff
@@ -1022,6 +1037,84 @@ def to_vector(self, transfer_attrs: bool = True) -> vector_module.GeoVectorDatas
...
+            warnings.warn(
+                f"Unknown {name} '{out}'. Data may not be processed correctly. "
+                f"Known {name}s are {supported}. Contact the pycontrails "
+                "developers if you believe this is an error."
+            )
+
+        return out
+
+    @property
+    def provider_attr(self) -> str:
```
mlshapiro, line 1069: What do you think about removing the `attr` suffix from these? 

It works well for `provider` and `product` but we'd probably want to rename `dataset` to `dataset_name` or something?

pycontrails/core/met.py
```diff
@@ -2389,9 +2484,12 @@ def originates_from_ecmwf(met: MetDataset | MetDataArray) -> bool:
     - :class:`HRES`
 
     """
-    return met.attrs.get("met_source") in ("ERA5", "HRES") or "ecmwf" in met.attrs.get(
-        "history", ""
-    )
+    if isinstance(met, MetDataset):
+        try:
+            return met.provider_attr == "ECMWF"
```
mlshapiro, line 2489: This is really nice - this motivates my comment above, `met.provider` would be a nice access pattern

mlshapiro: OK nevermind, the `_attr` suffix is probably good for clarity

pycontrails/core/met.py
mlshapiro: I take it back, I think the `_attr` suffix is probably worth ith

pycontrails/datalib/ecmwf/hres.py
```diff
@@ -241,15 +234,6 @@ class HRES(ECMWFAPI):
     #: Forecast run time, either specified or assigned by the closest previous forecast run
     forecast_time: datetime
 
-    #: User provided ``ECMWFService`` url
```
mlshapiro, line 244: Nice cleanup

pycontrails/models/accf.py
```diff
@@ -377,3 +375,30 @@ def _update_accf_config(self) -> None:
...
+            raise KeyError(msg) from e
+
+        if unit == "J m**-2":
+            continue
+        if unit != "W m**-2":
+            msg = f"Unexpected units '{unit}' for '{name}'. Expected 'J m**-2' or 'W m**-2'."
+            raise ValueError(msg)
+
+        # Convert from W m**-2 to J m**-2
+        da.attrs["units"] = "J m**-2"
```
mlshapiro, line 401: I think these two lines are here in error

pycontrails/models/accf.py
```diff
@@ -132,47 +154,34 @@ class ACCF(Model):
...
-        self._update_accf_config()
         if surface:
-            self.surface = surface.copy()
-
-        # This variable won't get used since we are not writing the output
-        # anywhere, but the library will complain if it's not defined
-        self.path_lib = "./"
+            surface = surface.copy()
+            surface = standardize_variables(surface, self.sur_variables)
+            surface.data = _rad_instantaneous_to_accumulated(surface.data)
```
mlshapiro, line 181: I think we can restrict ACCF usage to ECMWF ERA5 or HRES and then we can skip the `_rad_instantaneous..` step.

Or are you tryting to handle a different circumstnace here? The bug in #64 is due to us converting J/m2 to W/m2, not the other way around.

pycontrails/models/sac.py
```diff
@@ -142,24 +148,15 @@ def eval(
...
-
-        # In GeoVectorDataset case, return source
+        # Tag output with additional metadata attrs
+        self.transfer_met_source_attrs()
+        self.source.attrs["pycontrails_version"] = pycontrails.__version__
+        if scale_humidity:
+            for k, v in humidity_scaling.description.items():
+                self.source.attrs[f"humidity_scaling_{k}"] = v
+        if isinstance(engine_efficiency, (int, float)):
+            self.source.attrs["engine_efficiency"] = engine_efficiency
```
mlshapiro, line 158: should we put fuel on output since we're not copying it from source anywhere?

mlshapiro (commented):
I didn't look through examples or tests, but I gave my comments now to get started!

I agree with all your choices - a few naming discussion point, and then a potential typo or two, but it looks good!

pycontrails/core/met.py
zebengberg: Alternatively, if the `attrs` key is changed to `pycontrails_provider`, the property could also be `pycontrails_provider`.

pycontrails/models/accf.py
zebengberg: I think this is what I want ... but I might have my eyes crossed. I slightly changed this to 9878bec874ca2c8ee0e5175c856a121ad17b0d5e but that is only a syntax change.

If we end up on this line of code, that means the units are `W m**-2`. So I'm reassigning them to `J m**-2`, then multiplying by 3600 to make the conversion on the data itself.

pycontrails/models/accf.py
zebengberg: I think we probably should leave this `_rad_instantaneous_to_accumulated` function. If anyone has previously cached ERA5 or HRES radiation data with pycontrails (for example, us), then it will have radiation expressed in W/m2. And after we merge this branch, any newly cached data will have radiation expressed in J/m2. So this handles both cases by inspecting the unit.

pycontrails/models/sac.py
zebengberg: The main reason why I didn't is because our `Fuel` instance isn't harmonious with `self.source.data.to_netcdf(...)`. We'd have to handle it some other way ... like with how we handle `humidity_scaling` just above this for example.

There are a few other similar examples of this sort of thing kicking around in pycontrails. We have certain metadata that we keep in a dictionary (`Models.params` or `MetDataset.attrs` or `Flight.attrs`), and we want these to be serializable to JSON or netCDF or parquet. To do this right is probably a fairly invasive change.

pycontrails/core/datalib.py
zebengberg: c9c2957fc4659033c477a6745c5c7bc11f12a531

pycontrails/core/datalib.py
zebengberg: c9c2957fc4659033c477a6745c5c7bc11f12a531

pycontrails/core/met.py
zebengberg: I removed the call to `upper()`. We don't use that sort of pattern when working with aircraft ICAO codes, so there's no reason to allow case insensitivity here. I think keeping this case sensitive will help keep everything more explicit.

c95b2b5b2ecfe111193fc8128c7a3b6386f0d880

pycontrails/core/met.py
zebengberg: c95b2b5b2ecfe111193fc8128c7a3b6386f0d880

pycontrails/core/met.py
zebengberg: I feel mixed on this. I think requiring a "provider" key here is akin to us requiring an "aircraft_type" in the `Flight.attrs` on certain models. By analogy, renaming "aircraft_type" to "pycontrails_aircraft_type" would be fairly unwieldy ... we're already quite used to "aircraft_type" as a required `Flight.attrs` key. To me, this is similar in our syntax and usage.

pycontrails/core/met.py
zebengberg: I don't know! We could say that it'll be removed in the next minor version update ... does this sound right to you? (In this particular example, I think we _could_ just remove it right away.)

pycontrails/core/met.py
mlshapiro: I agree with you, feel free to leave as is!

pycontrails/models/accf.py
mlshapiro: Ah right, good point.

pycontrails/models/sac.py
mlshapiro: Ah ok, good point.

pycontrails/models/accf.py
mlshapiro: This was my mis-reading. Your new implementation is much clearer.

pycontrails/datalib/ecmwf/hres.py
```diff
@@ -635,47 +626,31 @@ def open_metdataset(
...
+            product = "ensemble"
+        else:
+            msg = f"Unknown stream type {self.stream}"
+            raise ValueError(msg)
+
+        ds.attrs.update(
+            provider="ECMWF",
+            dataset="HRES",
+            product=product,
+            hres_accumulated=True,
```
mlshapiro, line 651: I didn't notice this on the first pass - are there circumstances where `hres_accumulated` is not True?  We don't have this for ERA5 sources.

mlshapiro: Based on your comment in CoCiP it looks like this is the case, but just confirming.

mlshapiro: What about calling this `radiation_accumulated`?

trdean1:
I was just testing this out with ACCFs and found that there's no problem evaluating ACCFs over a flight (there is a test case that captures this), but I get an error message evaluating over a grid like [this](https://py.contrails.org/examples/ACCF.html#Grid-ACCF-with-Parameters). I end up with an error message like this: 

```
ValueError: conflicting sizes for dimension 'longitude': length 720 on 'olr' and length 1440 on {'longitude': 'air_temperature', 'latitude': 'air_temperature', 'level': 'air_temperature', 'time': 'air_temperature'}
```

pycontrails/datalib/ecmwf/hres.py
zebengberg: aeeacd39055bf8e400cdaf0eab8379ef3d5b2776

Any data coming from our `HRES` interface will have `hres_accumulated=True` because we swap the "step" and the "time" dimension in processing it. But it'd be possible to BYO for which this is not True.

pycontrails/models/accf.py
```diff
@@ -233,16 +245,15 @@ def eval(
                 self.params["horizontal_resolution"] = float(hres)
 
         self.set_source_met()
-        self._update_accf_config()
```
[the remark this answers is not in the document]
zebengberg: Thanks! I'll try to get this fixed.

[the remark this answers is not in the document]
zebengberg: Thanks for finding this bug.

I _think_ this is fixed in 26d797ad5b2dcd4e0fb48bacd48e3f47f92001da

Let me know if you see anything else weird in the `ACCF` interface. I refactored a few things (changed methods to functions, etc.) ... I'm happy to revert any of that.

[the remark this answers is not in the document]
trdean1: Awesome, this looks good, and it looks like the daytime contrail aCCF looks correct

PR #105, opened 2023-10-27
Feature/shift longitude reverse

mlshapiro:
## Changes

This is a tiny old branch for a small utility function. I added some test so we make sure to get this into the library.

#### Features

- Adds `met.shift_longitude_reverse` function to translate longitude coordinates from [180, 180) -> [0, 360) domain

## Tests

- [ ] CI tests pass

### DIFF
--- a/pycontrails/core/met.py
+++ b/pycontrails/core/met.py
@@ -2245,7 +2245,6 @@ def shift_longitude(data: XArrayType) -> XArrayType:
     data : XArrayType
         :class:`xr.Dataset` or :class:`xr.DataArray` with longitude dimension
 
-
     Returns
     -------
     XArrayType
@@ -2254,6 +2253,26 @@ def shift_longitude(data: XArrayType) -> XArrayType:
     return data.assign_coords(
         longitude=((data["longitude"].values + 180.0) % 360.0) - 180.0
     ).sortby("longitude", ascending=True)
+
+
+def shift_longitude_reverse(data: XArrayType) -> XArrayType:
+    """Shift longitude values from [-180, 180) to [0, 360) domain.
+
+    Sorts data by ascending longitude values.
+
+    Parameters
+    ----------
+    data : XArrayType
+        :class:`xr.Dataset` or :class:`xr.DataArray` with longitude dimension
+
+    Returns
+    -------
+    XArrayType
+        :class:`xr.Dataset` or :class:`xr.DataArray` with longitude values on [0, 360).
+    """
+    return data.assign_coords(longitude=(data["longitude"].values + 360.0) % 360.0).sortby(
+        "longitude", ascending=True
+    )
 
 
 def _wrap_longitude(data: XArrayType) -> XArrayType:
--- a/pycontrails/core/met.py
+++ b/pycontrails/core/met.py
@@ -2235,44 +2235,30 @@ def _is_zarr(ds: xr.Dataset | xr.DataArray) -> bool:
     return dask0.array.array.array.__class__.__name__ == "ZarrArrayWrapper"
 
 
-def shift_longitude(data: XArrayType) -> XArrayType:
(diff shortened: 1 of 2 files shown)


pycontrails/core/met.py
```diff
@@ -2248,6 +2247,26 @@ def shift_longitude(data: XArrayType) -> XArrayType:
     ).sortby("longitude", ascending=True)
 
 
+def shift_longitude_reverse(data: XArrayType) -> XArrayType:
```
zebengberg, line 2250: It's a little more work, but you could combine the two shift longitude functions into one. Something like this:

```python
def shift_longitude_more_general(data: XArrayType, a: float) -> XArrayType:
    return data.assign_coords(
        longitude=((data["longitude"].values + a) % 360.0) - a
    ).sortby("longitude", ascending=True)
```

This would shift the longitude from any domain to the interval `[-a, 360 - a)`. Then your new `shift_longitude_reverse` is the same as evaluating this one with `a=0.0`. And the current `shift_longitude` is the same as evaluating this one with `a=180.0`.

pycontrails/core/met.py
```diff
@@ -2248,6 +2247,26 @@ def shift_longitude(data: XArrayType) -> XArrayType:
...
+    ----------
+    data : XArrayType
+        :class:`xr.Dataset` or :class:`xr.DataArray` with longitude dimension
+
+    Returns
+    -------
+    XArrayType
+        :class:`xr.Dataset` or :class:`xr.DataArray` with longitude values on [0, 360).
+    """
+    return data.assign_coords(longitude=(data["longitude"].values + 360.0) % 360.0).sortby(
```
zebengberg, line 2265: If you don't want to use the more general setup I described above, I think the `+ 360.0` term is unnecessary here.

tests/unit/test_met.py
```diff
@@ -215,21 +215,56 @@ def test_wrap_longitude(zero_like_da: xr.DataArray, met_ecmwf_pl_path: str) -> N
 def test_shift_longitude(zero_like_da: xr.DataArray) -> None:
     da = zero_like_da.copy()
 
+    # assign random values to all coords
+    da.loc[:] = np.random.rand(*da.shape)  # noqa: NPY002
```
zebengberg, line 219: To avoid the [ruff NPY002 error](https://docs.astral.sh/ruff/rules/numpy-legacy-random/), you can do this:

```python
rng = np.random.default_rng()
da.loc[:] = rng.random(*da.shape)
```

It's an extra line of code, but I think it's better practice. If you want to reuse `rng` across different tests, you could define it as a fixture.

pycontrails/core/met.py
```diff
@@ -2248,6 +2247,26 @@ def shift_longitude(data: XArrayType) -> XArrayType:
     ).sortby("longitude", ascending=True)
 
 
+def shift_longitude_reverse(data: XArrayType) -> XArrayType:
+    """Shift longitude values from [-180, 180) to [0, 360) domain.
```
zebengberg, line 2251: I think both in this docstring and the one above, we can remove this part:

> Shift longitude values ~~from [-180, 180)~~ to [0, 360) domain.

The implementation works for any starting domain.

zebengberg (approved):
Thanks expanding our unit tests! Feel free ignore any of my suggestions.

pycontrails/core/met.py
mlshapiro: Fantastic suggestion, simplified and included in 4e7fbf3cea7942b56b4255ffdd4be4f607406c94

tests/unit/test_met.py
mlshapiro: 4e7fbf3cea7942b56b4255ffdd4be4f607406c94

pycontrails/core/met.py
mlshapiro: Removed

PR #139, opened 2024-01-22
Fix/windows interp

zebengberg:
Update cython annotations for `scipy` 1.12 compatibility.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,6 +39,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         pyversion: ['3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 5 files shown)


nickmasson (approved):
🚀 
Is there merit in writing some cython .pyx tests for these? 
And/Or do we have decent coverage for the behavior of these methods thru other unittests that touch this transitively?

zebengberg:
> 🚀 Is there merit in writing some cython .pyx tests for these? And/Or do we have decent coverage for the behavior of these methods thru other unittests that touch this transitively?

Yes -- everything in `rgi_cython.pyx` gets tested somewhat well through the `test_interpolation.py` module. Ideally our `rgi_cython.pyx` would itself live in scipy, but there are a few incompatibilities currently (we support float32 and float64 grid interpolation whereas scipy supports float64 and complex128).

PR #140, opened 2024-01-22
Feature/113

### LINKED ISSUE #113: Better interpolation bounds check in Cocip
Currently `Cocip` emits an [uninformative warning](https://github.com/contrailcirrus/pycontrails/blob/a5213f30fbd2d81a90738179199fe2974ef8e28a/pycontrails/models/cocip/contrail_properties.py#L447) if all tau contrail values are nan during contrail evolution. This should be removed or replaced with a more pointed warning describing the interpolation out-of-bounds issue.

zebengberg:
Closes #113 

#### Fixes

- Remove opaque warning issued when all tau_contrail values are nan in `Cocip` evolution.
- Emit warning in `Cocip.eval` if the advected contrail is blown outside of the domain of the met data.

## Tests

- [x] CI tests pass

### DIFF
--- a/pycontrails/models/cocip/cocip.py
+++ b/pycontrails/models/cocip/cocip.py
@@ -2185,6 +2185,12 @@ def calc_timestep_contrail_evolution(
         },
         copy=False,
     )
+    intersection = contrail_2.coords_intersect_met(met)
+    if not np.any(intersection):
+        warnings.warn(
+            f"At time {time_2}, the contrail has no intersection with the met data. "
+            "This is likely due to the contrail being advected outside the met domain."
+        )
 
     # Update cumulative radiative heating energy absorbed by the contrail
     # This will always be zero if radiative_heating_effects is not activated in cocip_params
--- a/pycontrails/models/cocip/cocip.py
+++ b/pycontrails/models/cocip/cocip.py
@@ -507,17 +507,19 @@ class Cocip(Model):
         if "waypoint" not in self.source:
             # Give flight a range index
             self.source["waypoint"] = np.arange(self.source.size)
-        elif not isinstance(self.source, Fleet):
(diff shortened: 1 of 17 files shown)


pycontrails/models/cocip/cocip.py
```diff
@@ -2185,6 +2185,12 @@ def calc_timestep_contrail_evolution(
         },
         copy=False,
     )
+    intersection = contrail_2.coords_intersect_met(met)
+    if not np.any(intersection):
+        warnings.warn(
```
nickmasson, line 2190: is this something that the developer implementing pycontrails can manage/control? e.g. can they do something differently with their met data and problem definition/configuration as to avoid having a contrail advect out of bounds?

Or is this more of a runtime and circumstantial issue, where we simply want to let the developer (and any consumers of the outputs) know that these data may be troubled due to a truncated evolution (but the implementer has no agency over remediating the issue)?

If the latter, I wonder if this warrants a python logged warning, rather than python native (stderr) warning?

nickmasson (approved):
lgtm from what I can tel!

Do we care if the tests for Windows machines pass? looks like they're failing CICD testing.

zebengberg:
> Do we care if the tests for Windows machines pass? looks like they're failing CICD testing.

Yep, we do care! They're passing now after I rebased off my other branch with the fix.

pycontrails/models/cocip/cocip.py
zebengberg: I think this is more along the lines of the first. If the contrail advects of out of bounds, the developer ideally should have used more met data in running the model.

Occasionally the developer could have done everything correct and still the contrails blows away ... so it's a little debatable here. But I do think the essence of this warning is that the developer could have done something slightly better / different.

PR #141, opened 2024-01-23
Feature/better filtering

### LINKED ISSUE #51: Resample, fill, and filter flight trajectories
**Description**

We want to provide an opinionated, general set of methods for preparing flight trajectories for analysis.

Generally this includes:

- Resample trajectories to have a waypoint every ~ 10s - 1 min
- Interpolate spatial (lat, lon) gaps when upsampling. Interpolate small gaps linearly, interpolate longer gaps with geodesics.
- Filter the altitude signal to have an reasonable flight profile with distinct flight phases

Currently we have implemented `Flight.resample_and_fill` which works pretty well. This function contains some logic to validate an vertical profile, but the logic may be out of date. 

- [ ] If possible, make `resample_and_fill` a module function (`flight.resample_and_fill`) and call the module function from the class.

We also have a function [`flight.filter_altitude`](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/core/flight.py#L1439) that should remove high frequency noise (usually ADS-B or FDR)  and snap cruise altitudes to 1000s of feet. This should be a cascade of filters with a (1) median filter (2) low pass filter (3) cruise altitude correction. Feel free to design a better methodology that applies to all the possible segment durations we may encounter.

- [x] Take `segment_duration` as an input to `filter_altitude`. In the class method version, we can calculate segment duration automatically
- [x] Create a class method `Flight.filter_altitude` that calls the module function
- [x] Design a set of filtering strategies that work with the set of altitude/segment duration inputs we may receive. Ideally we'll get data with a time interval of < 10s minutes. We may decide we can't filter higher interval data, so we can throw an error if the segment duration > 30 min or whatever starts to really fail.
- [x] Implement filtering strategy
- [x] Implement tests for (1) flight with very high frequency altitude data (segment_duration ~ 1s) (2) nominal frequency (segment_duration ~ 1min) (3) lower frequency (segment_duration ~ 10 min) (3) low frequency (like an ATC route string)

As a final step, confirm that the implementations of `resample_and_fill` and `filter_altitude` play well together. In the `Flight` class methods, we can make sure this is handled well. In the module functions, we can reference the input sources (e.g. `flight.segment_duration`) and suggest the best way to assemble cleaned trajectories.

- [x] Update the [Flight.ipynb](https://github.com/contrailcirrus/pycontrails/blob/main/docs/examples/Flight.ipynb) with an example of cleaning a trajectory

**Alternatives**

- Use the [traffic.core.flight.Flight.filter](https://traffic-viz.github.io/api_reference/traffic.core.flight.html#traffic.core.Flight.filter). This relies on the client side implementing the cascading filter. I'd prefer to implement an opinionated set of fliters that work best, and then raise warning/errors when we don't get good results

**Additional context**

- ~1s frequency (noisy) altitude measurement. [tests/unit/static/flight-spire-data-cleaning.pq](https://github.com/contrailcirrus/pycontrails/blob/main/tests/unit/static/flight-spire-data-cleaning.pq)
- ~1min frequency [tests/unit/static/flight.csv](https://github.com/contrailcirrus/pycontrails/blob/main/tests/unit/static/flight.csv)
- 1 min frequency: [tests/unit/static/flight-cocip2.csv](https://github.com/contrailcirrus/pycontrails/blob/main/tests/unit/static/flight-cocip2.csv)

trdean1:
Closes #51 

## Changes

> Removes `fit_altitude` method in favor of `filter_altitude`
> Reworks `filter_altitude` to only apply median filter during cruise
> Creates `clean_and_resample` class method
> Creates `filter_altitude` class method

## Internals

The `clean_and_resample` method uses the following logic:
> For flights with a median sample rate >10 seconds, no smoothing is applied
> For flights with gaps larger than 5 minutes, the flight is first resampled to a one second interval, then smoothed, then resampled to the requested rate

## Tests

- [x] CI tests pass


trdean1:
TODO
- [ ] Review failing test cases and implement more robust testing
- [ ] Create class method for `filter_altitude`
- [ ] Decide on interface for "automatic" cleaning/resampling

pycontrails/core/flight.py
```diff
@@ -1753,21 +1691,59 @@ def filter_altitude(
...
+    start_cruise = np.empty(cruise_duration.size, dtype=bool)
+    # if first sample is in cruise, first sample of start_cruise marks start of a segment
+    start_cruise[0] = ~not_cruise[0]
+    start_cruise[1:] = not_cruise[:-1] & ~not_cruise[1:]
+
+    start_idxs = np.flatnonzero(start_cruise)
+    end_idxs = np.flatnonzero(end_cruise)
+
+    # Threshold for min cruise segment
+    long_mask = cruise_duration[end_idxs] > 120
```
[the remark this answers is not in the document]
trdean1: Good catch - thanks!

pycontrails/core/fleet.py
```diff
@@ -341,19 +341,6 @@ def segment_angle(self) -> tuple[npt.NDArray[np.float_], npt.NDArray[np.float_]]
         cos_a[self.final_waypoints] = np.nan
         return sin_a, cos_a
 
-    @overrides
-    def fit_altitude(
```
zebengberg, line 345: I think you should keep this, but change the method name to `clean_and_resample`. If we call this new method on on a `Fleet`, we want some sort of error.

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
-        r2_target: float = 0.999,
-        max_cruise_rocd: float = 10.0,
-        sg_window: int = 7,
-        sg_polyorder: int = 1,
+        freq: str = "1min",
+        fill_method: str = "geodesic",
+        geodesic_threshold: float = 100e3,
+        nominal_rocd: float = constants.nominal_rocd,
+        kernel_size: int = 17,
+        cruise_threshold: float = 120,
```
zebengberg, line 948: 120 -> 120.0

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+            Threshold for geodesic interpolation, [:math:`m`].
+            If the distance between consecutive waypoints is under this threshold,
+            values are interpolated linearly.
+        nominal_rocd : float | None, optional
+            Nominal rate of climb / descent for aircraft type.
+            Defaults to :attr:`constants.nominal_rocd`.
+        kernel_size : int, optional
+            Passed directly to :func:`scipy.signal.medfilt`, by default 11.
+            Passed also to :func:`scipy.signal.medfilt`
+        cruise_theshold : int, optional
```
zebengberg, line 980: int -> float

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+        fill_method : {"geodesic", "linear"}, optional
+            Choose between ``"geodesic"`` and ``"linear"``, by default ``"geodesic"``.
+            In geodesic mode, large gaps between waypoints are filled with geodesic
+            interpolation and small gaps are filled with linear interpolation. In linear
+            mode, all gaps are filled with linear interpolation.
+        geodesic_threshold : float, optional
+            Threshold for geodesic interpolation, [:math:`m`].
+            If the distance between consecutive waypoints is under this threshold,
+            values are interpolated linearly.
+        nominal_rocd : float | None, optional
```
zebengberg, line 974: float | None -> float

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+            Nominal rate of climb / descent for aircraft type.
+            Defaults to :attr:`constants.nominal_rocd`.
+        kernel_size : int, optional
+            Passed directly to :func:`scipy.signal.medfilt`, by default 11.
+            Passed also to :func:`scipy.signal.medfilt`
+        cruise_theshold : int, optional
+            Minimal length of time, in seconds, for a flight to be in cruise to apply median filter
+        force_filter: bool, optional
+            If set to true, `filter_altitude` will always be called. otherwise, it will only
```
zebengberg, line 983:
```
`filter_altitude` -> :meth:`filter_altitude`
```

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
-            self.altitude_ft,
-            max_segments=max_segments,
-            pop=pop,
-            r2_target=r2_target,
-            max_cruise_rocd=max_cruise_rocd,
-            sg_window=sg_window,
-            sg_polyorder=sg_polyorder,
+        clean_flight: Flight
+        # If the flight has a large sampling period, don't try to smooth it unless requested
+        median_gap = np.nanmedian(self.segment_duration())
```
zebengberg, line 1005: You use the segment duration both here and down below when you calculate `max_gap`. I'd suggest storing it as a separate variable above this, then using it here and below.

```python
seg_duration = self.segment_duration()
median_gap = np.nanmedian(seg_duration)
...
max_gap = np.max(seg_duration)
```

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+                fill_method,
+                geodesic_threshold,
+                nominal_rocd,
+                drop,
+                keep_original_index,
+                climb_descend_at_end,
+            )
+
+        # If the flight has large gap(s), then call resample and fill, then filter altitude
+        max_gap = max(self.segment_duration())
```
zebengberg, line 1018: max -> np.max

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+        if max_gap > 300:
```
zebengberg, line 1019: 300 -> 300.0

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+            climb_descend_at_end,
         )
 
-        out = self.copy()
-        out.update(altitude_ft=alt_ft)
-        return out
+    def filter_altitude(
+        self,
+        kernel_size: int = 17,
+        cruise_threshold: float = 120,
```
zebengberg, line 1050: 120 -> 120.0

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+        Currently runs altitude through a median filter using :func:`scipy.signal.medfilt`
+        with ``kernel_size``, then a Savitzky-Golay filter to filter noise. The median filter
+        is only applied during cruise segments that are longer than ``cruise_threshold``.
+
+        Parameters
+        ----------
+        kernel_size : int, optional
+            Passed directly to :func:`scipy.signal.medfilt`, by default 11.
+            Passed also to :func:`scipy.signal.medfilt`
+        cruise_theshold : int, optional
```
zebengberg, line 1064: int -> float

pycontrails/core/flight.py
```diff
@@ -938,65 +938,165 @@ def resample_and_fill(
...
+        df = self.dataframe.copy()
+        if "altitude_ft" not in df:
+            df["altitude_ft"] = units.m_to_ft(df["altitude"])
+
+        altitude_filtered = filter_altitude(
+            df["time"], df["altitude_ft"], kernel_size, cruise_threshold
+        )
+        df["altitude_ft"] = altitude_filtered
+        if "altitude" in df:
+            df["altitude"] = units.ft_to_m(altitude_filtered)
```
zebengberg, line 1097: What about avoiding the dataframe entirely?

```python
        out = fl.copy()
        altitude_ft_filtered = filter_altitude(
            fl["time"], fl.altitude_ft, kernel_size, cruise_threshold
        )
        out.update(altitude_ft=altitude_ft_filtered)
        out.data.pop("altitude", None)  # avoid any ambiguity
        out.data.pop("level", None)  # avoid any ambiguity
        return out
```

pycontrails/core/flight.py
```diff
@@ -1740,34 +1838,73 @@ def filter_altitude(
...
+    # Find cruise phase in filtered profile
+    seg_duration = segment_duration(time)
+    seg_rocd = segment_rocd(seg_duration, altitude_filt)
+    seg_phase = segment_phase(seg_rocd, altitude_filt)
+    is_cruise = seg_phase == FlightPhase.CRUISE
+
+    # Compute cumulative segment time in cruise segments
+    v = np.nan_to_num(seg_duration)
+    v[~is_cruise] = 0
+    n = v == 0
```
zebengberg, line 1872: 0 -> 0.0

pycontrails/core/flight.py
```diff
@@ -1740,34 +1838,73 @@ def filter_altitude(
...
+    c = np.cumsum(v)
+    d = np.diff(c[n], prepend=0.0)
+    v[n] = -d
+    cruise_duration = np.cumsum(v)
+
+    # Find cruise segment start and end indices
+    not_cruise = cruise_duration == 0
```
zebengberg, line 1879: 0 -> 0.0

tests/unit/test_cocip.py
```diff
@@ -1132,6 +1132,7 @@ def test_cocip_gfs(fl: Flight, met_gfs: MetDataset, rad_gfs: MetDataset) -> None
     # assert np.nansum(out["ef"]) > 0
 
 
+@pytest.mark.filterwarnings("ignore:.*the contrail has no intersection with the met")
```
zebengberg, line 1135: Thank you!

zebengberg (approved):
Thanks Tom -- this looks great.

I left a bunch of nitty comments ... nothing critical.

PR #146, opened 2024-01-31
Pin black and ruff versions

### LINKED ISSUE #145: pin version of Black/Ruff
**Description**

Black and Ruff are used as part of code development control. They are not part of the critical path for the pycontrails application code.

It is a common occurrence that updated Black releases will cause issues (either causing previously passing code to fail, and/or, creating inconsistency between a local development environment and CICD environment.

As such, it is sensible to pin the version of these two packages to fixes versions, resulting in deterministic behavior across all environments.

**Task(s)**
Update all areas where Black/Ruff versions are declared to use a fixed/pinned version. This includes, but is not limited to, directives that set up CICD environment and local environment. If feasible, refactor to include a single reference for black/ruff version across the entire project (i.e. avoid declarations in multiple places).

**Acceptance Criteria**
Pycontrails in a production/installed environment, development environment and CICD environment all use the same version of black/ruff.

thabbott:
Closes #145

## Changes

Modify [pyproject.toml](https://github.com/contrailcirrus/pycontrails/compare/enhancement/pin_black_ruff?expand=1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) to pin `black` version (currently 24.1.1) and `ruff` version (currently 0.1.15), and add a `black` configuration argument so that `black` errors if it doesn't match the pinned version when run. AFAIK this is the only file that needs modification.

The changes don't quite meet the goal of avoiding duplicate version declarations (we're limited by the TOML file spec), but a mismatch between the `black` version pinned in development dependencies and the required version specified in `tools.black` should show up early in tests.

`ruff` doesn't have a flag that checks for a required version, so these changes won't prevent somebody from using a manually-up/downgraded version.

#### Internals

- Pin `black` and `ruff` versions for consistency between local and CI/CD environments

## Tests

- [ ] CI tests pass

## Reviewer

> @trdean1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ complete = ["pycontrails[ecmwf,gcp,gfs,goes,jupyter,pyproj,vis,zarr]"]
 
 # Development dependencies
 dev = [
-    "black[jupyter]>=23.12",
+    "black[jupyter]==24.1.1",
     "dep_license",
     "fastparquet>=0.8",
     "ipdb>=0.13",
@@ -72,7 +72,7 @@ dev = [
     "pytest>=6.1",
     "pytest-cov>=2.11",
     "requests>=2.25",
-    "ruff>=0.0.259",
+    "ruff==0.1.15",
 ]
 
 # Documentation / Sphinx dependencies
@@ -195,6 +195,7 @@ show_error_codes = true
 [tool.black]
 line-length = 100
 preview = true
+required_version = "24.1.1"
 
 # pytest
 # https://docs.pytest.org/en/7.1.x/reference/customize.html
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,10 +3,6 @@
 
 name: Unit tests
 
-defaults:
-  run:
-    shell: bash
-
 on:
   # run on push to main
   push:
@@ -119,6 +115,7 @@ jobs:
           pip list
 
       - name: QC & Test
+        shell: bash
         run: |
           pip check
           black pycontrails --check
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,6 +3,10 @@
 
 name: Unit tests
 
+defaults:
+  run:
+    shell: bash
+
 on:
   # run on push to main
   push:
@@ -116,7 +120,6 @@ jobs:
 
       - name: QC & Test
         run: |
-          set -e
           pip check
           black pycontrails --check
           ruff pycontrails tests
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -86,8 +86,7 @@ jobs:
 
       - name: Install pycontrails (dev)
         run: |
-          pip install -U pip wheel
-          pip install -e .[complete,dev,docs] --verbose
+          make dev-install
 
       # In latest-windows, redirecting stdout to a file uses utf-16 encoding
       # This gives an error when ssh tries to read the key
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -79,6 +79,11 @@ jobs:
           gcloud storage cp -r gs://contrails-301217-bada/bada/bada4 ${{ env.BADA_CACHE_DIR }}
           ls -l ${{ env.BADA_CACHE_DIR }}
 
+      - name: Install make - windows
+        if: ${{ matrix.os == 'windows-latest' }}
+        run: |
+          choco install make
+
       - name: Install pycontrails (dev)
         run: |
           pip install -U pip wheel
@@ -118,7 +123,7 @@ jobs:
         shell: bash
         run: |
           pip check
-          black pycontrails --check
-          ruff pycontrails tests
-          mypy pycontrails
-          pytest tests/unit -vv
+          make black-check
+          make ruff
+          make mypy
+          make pytest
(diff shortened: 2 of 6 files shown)


trdean1:
Awesome, yes I think this should do it!

Maybe a question for @zebengberg  - is there a reason why the check in the [action](https://github.com/contrailcirrus/pycontrails/blob/98efabf342054b27f72e603a9f28208140da0266/.github/workflows/test.yaml#L120) is different than the check in [the makefile](https://github.com/contrailcirrus/pycontrails/blob/bfae7745ea142340e3a1c0cbbf066ae4579b4e57/Makefile#L74) when it comes to what black gets run over?  

I think this is why Tristian had to update the two test files here, but those did not get flagged by the CI/CD tests when I made my changes

trdean1 (approved):

zebengberg:
> Awesome, yes I think this should do it!
> 
> Maybe a question for @zebengberg - is there a reason why the check in the [action](https://github.com/contrailcirrus/pycontrails/blob/98efabf342054b27f72e603a9f28208140da0266/.github/workflows/test.yaml#L120) is different than the check in [the makefile](https://github.com/contrailcirrus/pycontrails/blob/bfae7745ea142340e3a1c0cbbf066ae4579b4e57/Makefile#L74) when it comes to what black gets run over?
> 
> I think this is why Tristian had to update the two test files here, but those did not get flagged by the CI/CD tests when I made my changes

I think it's a historical artifact. We should try to rectify this.

pyproject.toml
```diff
@@ -195,6 +195,7 @@ show_error_codes = true
 [tool.black]
 line-length = 100
 preview = true
+required_version = "24.1.1"
```
zebengberg, line 198: Is this necessary? If someone is doing local pycontrails development, they should be using the dev dependencies ... should I think of this as an extra guardrail?

zebengberg (approved):

pyproject.toml
mlshapiro: Yes - I think this is. This will ensure that as this version is bumped, the team gets alerted to re-run the dev-install to bump black.

mlshapiro:
> > Awesome, yes I think this should do it!
> > Maybe a question for @zebengberg - is there a reason why the check in the [action](https://github.com/contrailcirrus/pycontrails/blob/98efabf342054b27f72e603a9f28208140da0266/.github/workflows/test.yaml#L120) is different than the check in [the makefile](https://github.com/contrailcirrus/pycontrails/blob/bfae7745ea142340e3a1c0cbbf066ae4579b4e57/Makefile#L74) when it comes to what black gets run over?
> > I think this is why Tristian had to update the two test files here, but those did not get flagged by the CI/CD tests when I made my changes
> 
> I think it's an historical artifact. We should try to rectify this.
> 
> I'm not sure if we can run Makefile recipes in the windows workflows, so addressing this may not be as simple as copying the recipe.

I can attest that this is true. Unfortunately we can't run the make on windows, so we'll either have to run the commands in CI, or just run the commands on the windows specific install.


Havin the windows specific test is important, so unfortunately I think the only thing to do is copy the commands into the action.

@thabbott Feel free to go ahead with this in this in the PR

pyproject.toml
thabbott: Unfortunately I don't think there's a similar guardrail for ruff---so somebody with an old dev environment could continue using a version of ruff that doesn't match the one specified in pyproject.toml.

thabbott:
> > Awesome, yes I think this should do it!
> > Maybe a question for @zebengberg - is there a reason why the check in the [action](https://github.com/contrailcirrus/pycontrails/blob/98efabf342054b27f72e603a9f28208140da0266/.github/workflows/test.yaml#L120) is different than the check in [the makefile](https://github.com/contrailcirrus/pycontrails/blob/bfae7745ea142340e3a1c0cbbf066ae4579b4e57/Makefile#L74) when it comes to what black gets run over?
> > I think this is why Tristian had to update the two test files here, but those did not get flagged by the CI/CD tests when I made my changes
> 
> I think it's an historical artifact. We should try to rectify this.
> 
> I'm not sure if we can run Makefile recipes in the windows workflows, so addressing this may not be as simple as copying the recipe.

I'm a little confused what this has to do with Makefile recipes in Windows workflows. Is the idea that we'd rather be able to run `make black-check` in the GitHub action (so we don't have to manually maintain agreement with the Makefile), but can't because we don't currently install `make` in Windows workflows?

thabbott:
I think I've taken care of the windows workflow issues: [test.yaml](https://github.com/contrailcirrus/pycontrails/pull/146/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692) now
- includes a step to install `make` on windows runners
- uses Makefile recipes to installs a pycontrails dev environment and lint/test
- specifies `bash` as the shell for the lint/test step, which stops windows workflows from passing despite linting errors (compare https://github.com/contrailcirrus/pycontrails/pull/146/commits/6dc660d679f77dc9233e4a177cf513318ad1972a, where windows runners should fail but don't, to https://github.com/contrailcirrus/pycontrails/pull/146/commits/dc06df164dd3597ed09acb9a054484512b4dd156, where they correctly fail)
Any outstanding issues I should address before I merge?

zebengberg:
> Can I use a Makefile recipe to install pycontrails-bada, or is should I leave those steps as they are?

I think leave as is! I don't think this recipe would work for windows, even with your support for make here.

PR #147, opened 2024-01-31
Improvements to develop instructions

### LINKED ISSUE #142: Minor pain points in develop instructions
I ran into a few minor pain points while following instructions for creating a pycontrails development environment (https://py.contrails.org/develop.html). I'm documenting them here in case the record is useful. Addressing them might somewhat improve the new-developer experience, but probably shouldn't be a priority.

- "Copy to clipboard" buttons copy a leading `$`. (This is true throughout the online documentation.) 
- I couldn't clone the pycontrails repository via SSH, probably because I don't have an SSH key set up. git's error message isn't very informative, though, so it may be worth adding a note about setting up SSH keys (https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#adding-your-ssh-key-to-the-ssh-agent) or about cloning via HTTPS.
- Some tests require a CDS API key (https://cds.climate.copernicus.eu/api-how-to) and/or access to GCP storage buckets. Possibly worth noting these requirements so users aren't surprised when some tests fail.

thabbott:
Closes #142 

I'm opening this as a draft PR because I'd like some feedback on potential changes to the Makefile testing harness. 

The changes address testing issues mentioned in #142 by setting variables in recipes that check for GCP credentials
https://github.com/contrailcirrus/pycontrails/blob/d56a43d5bd6d70c988b99df641ea499b560e1c81/Makefile#L138-L139
and ensure required ERA5 data is cached
https://github.com/contrailcirrus/pycontrails/blob/d56a43d5bd6d70c988b99df641ea499b560e1c81/Makefile#L142-L152
These variables are used to skip notebook tests (`make nb-test`) if cached data isn't available
https://github.com/contrailcirrus/pycontrails/blob/d56a43d5bd6d70c988b99df641ea499b560e1c81/Makefile#L209-L219
and to skip doctests (`make doctest`) if cached data and/or GCP credentials are missing
https://github.com/contrailcirrus/pycontrails/blob/d56a43d5bd6d70c988b99df641ea499b560e1c81/Makefile#L167-L176

These changes prevent the Makefile-based testing harness from breaking for users withou CDS and/or GCP credentials. Instead, users are notified (in yellow) that notebook tests or doctests were skipped because of missing credentials. I can add a note to the develop instructions noting that this might happen, with some clues about how to get the tests to run.

The changes feel a bit hacky to me (and break Makefile syntax highlighting in my editor and on GitHub, which is annoying), and the conditional skipping is all-or-nothing (e.g., either all the notebook tests run, or none do). I haven't been able to think of a better solution, though. So, we could
1. Conditionally skip notebook tests and doctests at the level of the Makefile (my current approach)
2. Maybe use some other switch to conditionally skip notebook tests and doctests (I haven't found one)
3. Leave the Makefile as is and add a note to the documentation warning users that parts of the testing harness will break without CDS and GCP credentials.

Any thoughts/preferences from others?


zebengberg (approved):
Thanks Tristan! I can confirm that all of your changes work for me locally.

Makefile
mlshapiro, line 1: I don't think this is the right implementation - lets talk through this tomorrow!

mlshapiro (commented):
I think we should handle this all within the tests themselves, rather than in the Makefile.

I think its a good idea to a comment to the develop docs that some of the tests may break without CDS or GCP credentials. So perhaps just update the docs in this PR, and then lets pick up proper handling of the failing tests separately!

The other docs changes look great! Thanks

thabbott:
Apparently sympy (at least as of 2018) used some custom decorators to conditionally skip doctests---see https://github.com/scientific-python/pytest-doctestplus/issues/16. This seems like a much nicer alternative for doctests, but I don't think it's a solution for conditionally skipping *notebook* tests.

Makefile
```diff
@@ -9,6 +9,10 @@ help:
 
 .PHONY: help
 
+# Colors
+COLOR_YELLOW = \033[0;33m
```
mlshapiro, line 13: Really glad to learn about this!

Makefile
```diff
@@ -129,17 +133,8 @@ main-test-status:
 DOCS_DIR = docs
 DOCS_BUILD_DIR = docs/_build
 
-# Common ERA5 data for nb-tests and doctests
+# Cache common ERA5 data for nb-tests and doctests in DiskCacheStore(cache_dir=".doc-test-cache")
 ensure-era5-cached:
-	python -c 'from pycontrails.datalib.ecmwf import ERA5; \
```
mlshapiro, line 134: Consolidate `cache-era5-gcp` and `ensure-era5-cached` into one call with a common DiskCacheStore

Makefile
```diff
@@ -148,11 +143,18 @@ cache-era5-gcp:
 		met_vars = ["t", "q", "u", "v", "w", "ciwc", "z", "cc"]; \
 		rad_vars = ["tsr", "ttr"]; \
 		ERA5(time=time, variables=met_vars, pressure_levels=lev, cachestore=cache).download(); \
-		ERA5(time=time, variables=rad_vars, cachestore=cache).download()'
-
+		ERA5(time=time, variables=rad_vars, cachestore=cache).download()' \
+		|| ( \
+		echo -e \
+			"$(COLOR_YELLOW)CDS Credentials required to run notebooks and doc tests\n" \
```
mlshapiro, line 149: Throw the error once here so it breaks out of the tests with a helpful message? Maybe there is a better URL to use here

mlshapiro (commented):
@thabbott I slimmed down the Makefile changes here - what do you think?

PR #150, opened 2024-02-01
Remove empty flight in Fleet.from_seq

### LINKED ISSUE #149: Error when creating `Fleet` from sequence with empty `Flight`s
Opening an issue to note an error that I found difficult to debug. I'll assign myself and work on an improvement.

The error appears when creating a `Fleet` from a sequence of `Flight`s, some of which are missing data:
```python
import pandas as pd
import numpy as np
from pycontrails import Flight, Fleet

df = pd.DataFrame()
df["longitude"] = np.linspace(0, 50, 100)
df["latitude"] = np.linspace(0, 10, 100)
df["altitude"] = 11000
df["time"] = pd.date_range("2022-03-01 00:00:00", "2022-03-01 02:00:00", periods=100)

flights = [Flight(df, flight_id=1), Flight(flight_id=2)]
Fleet.from_seq(flights)
```
results in `ValueError: Unexpected flight_id(s) {2} in fl_attrs.` This simple example is not too hard to debug, but I ran into this error when `resample_and_fill` dropped a small fraction of a large number of flights to 0 length, and it took me a while to figure out what the problem was. I think we probably want to catch this case earlier on and provide a more informative error.

zebengberg:
Closes #149 

#### Fixes

- Remove empty flights in `Fleet.from_seq`.

## Tests

- [ ] CI tests pass

### DIFF
--- a/pycontrails/core/fleet.py
+++ b/pycontrails/core/fleet.py
@@ -256,12 +256,12 @@ class Fleet(Flight):
 
     def segment_true_airspeed(
         self,
-        u_wind: npt.NDArray[np.float_] | float = 0.0,
-        v_wind: npt.NDArray[np.float_] | float = 0.0,
+        u_wind: npt.NDArray[np.float64] | float = 0.0,
+        v_wind: npt.NDArray[np.float64] | float = 0.0,
         smooth: bool = True,
         window_length: int = 7,
         polyorder: int = 1,
-    ) -> npt.NDArray[np.float_]:
+    ) -> npt.NDArray[np.float64]:
         """Calculate the true airspeed [:math:`m / s`] from the ground speed and horizontal winds.
 
         Because Flight.segment_true_airspeed uses a smoothing pattern, waypoints in :attr:`data`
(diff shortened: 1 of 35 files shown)


pycontrails/core/fleet.py
```diff
@@ -174,10 +178,11 @@ def from_seq(
...
         """
-        if copy:
-            seq = tuple(fl.copy() for fl in seq)
-        elif not isinstance(seq, (list, tuple)):
-            seq = tuple(seq)
+
+        def _maybe_copy(fl: Flight) -> Flight:
+            return fl.copy() if copy else fl
+
+        seq = tuple(_maybe_copy(fl) for fl in seq if fl)  # filter out empty flights
```
thabbott, line 185: Is it worth emitting a warning if some empty flights are filtered out? I'm not sure why you would intentionally pass in an empty flight, so this might be a signal that something has gone wrong in an earlier preprocessing step.

thabbott (approved):
Looks good to me---thanks for taking care of this! Just left one comment asking whether we want to emit a warning when we catch this corner case.

pycontrails/core/fleet.py
zebengberg: @thabbott Yeah, I thought about this exact question, and I wasn't sure! I'll completely defer to you here. You encountered this issue in a notebook in which you called `Fleet.resample_and_fill`. Would you have been happier with a warning? I think a warning makes sense if there's something you could have done differently -- should you have previously removed the short flights?

zebengberg: e2871f6db3fb267169f425d9ff7d4ac4cee88292

Issue #153, opened 2024-02-14
UserWarning: Met data 'rad' does not overlap the grid domain along the time axis. Lead to missing values in first hour of eval.


kirscim:
**Description**

Loading met and rad data from cachestore with same start and end times leads to UserWarning only for rad parameters. Consequently, the first hour of eval does not contain any values. I think this is not the intended behavior, since ::func:_check_met_rad_times (lines 141-153 in cocip_time_handling) checks rad and met with same tmin and tmax values.

**Details**

- Version: v0.49.3 (but also v0.49.2/1)
- Module: models.cocipgrid.cocip_time_handling
- OS: linux cluster
- Reporter: kirscsim

**Steps to Reproduce**

Minimal example:

```import numpy as np

from pycontrails import DiskCacheStore
from pycontrails.datalib.ecmwf import ERA5, HRES
from pycontrails.models.cocip import Cocip
from pycontrails.models.cocipgrid import CocipGrid
from pycontrails.models.ps_model.ps_grid import PSGrid
from pycontrails.core.met import MetDataset
from pycontrails.models.humidity_scaling import ConstantHumidityScaling
from datetime import datetime, timedelta
```

Data loading from cachestore and preprocessing:

```
t0 = datetime(2024,2,5,0)
t1 = datetime(2024,2,5,13)

pressure_levels = [
    900, 850, 800, 700, 600, 500, 400, 300, 250, 200, 150, 100, 70, 50, 30
]

forecast_time = datetime(2024,2,5,0)

cachestore = DiskCacheStore(
        cache_dir='path/to/cachestore', 
        allow_clear=False,
    )

met = HRES(
        time=(t0, t1),
        variables=Cocip.met_variables + Cocip.optional_met_variables, 
        pressure_levels=pressure_levels, 
        grid=.25,
        cachestore=cachestore,
        url='https://api.ecmwf.int/v1',
        key=***,
        email=***,
        forecast_time=forecast_time,
)
    
rad = HRES(
    time=(t0, t1),
    variables=Cocip.rad_variables,
    grid=.25,
    cachestore=cachestore, 
    url='https://api.ecmwf.int/v1',
    key=***,
    email=***,
    forecast_time=forecast_time,
)

met = met.open_metdataset(xr_kwargs=dict(parallel=False))
rad = rad.open_metdataset(xr_kwargs=dict(parallel=False))

met = met.wrap_longitude()
rad = rad.wrap_longitude()
```

Met times:
[...]

PR #155, opened 2024-02-23
Fix/cocip grid met time

zebengberg:
## Changes

Support running `CocipGrid` on meteorology data without a uniformly-spaced time dimension.

#### Breaking changes

- Remove the `CocipGridParams.met_slice_dt` parameter. Now met downselection is handled automatically during contrail evolution. When the `met` and `rad` data passed into `CocipGrid` are not already loaded into memory, this update may make `CocipGrid` slightly more performant.
- No longer explicitly load `met` and `rad` time slices into memory in `CocipGrid`. This now only occurs downstream when interpolation is performed. This change better aligns `CocipGrid` with other pycontrails models.
- Remove the `cocipgrid.cocip_time_handling` module. Any useful tooling has been moved directly to the `cocipgrid.cocip_grid` module.
- Remove the `CocipGrid.timedict` attribute. Add a `CocipGrid.timesteps` attribute. This is now applied in the same manner that the `Cocip` model uses its `timesteps` attribute.
- Simplify the runtime estimate used in constructing the `CocipGrid` `tqdm` progress bar. The new estimate is less precise than the previous estimate and should not be trusted for long-running simulations.

#### Features

- Add a `MetDataset.downselect_met` method. This performs a met downselection in analogy with `GeoVectorDataset.downselect_met`.

## Tests

- [ ] CI tests pass

## Reviewer

@thabbott 
@tsankar


tests/unit/test_cocip_grid.py
```diff
@@ -773,3 +741,14 @@ def test_verbose_outputs_formation(
     assert ds["fuel_flow"].mean() == pytest.approx(0.6037, rel=rel)
     assert ds["rhi"].mean() == pytest.approx(0.6273, rel=rel)
     assert ds["iwc"].mean() == pytest.approx(4.4621e-06, rel=rel)
+
+
+def test_max_age_exceeds_met(instance_params: dict[str, Any], source: MetDataset) -> None:
+    """Perform smoke test to ensure max_age can exceed the available met time."""
```
tsankar, line 747: For my understanding, what actually happens in this case? Is the met data extended somehow, or do you fall back to the last available time?

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -212,47 +200,105 @@ def eval(
...
+                        contrail_list.append(res.contrail)
+                if self.params["verbose_outputs_formation"] and res.verbose_dict:
+                    verbose_dicts.append(res.verbose_dict)
 
-            vectors, summary_data, verbose_dicts_this_step, contrail_lists = zip(*evolved)
-            existing_vectors = combine_vectors(vectors, self.params["target_split_size"])
-            summary = VectorDataset.sum([r for r in summary_data if r])
-            summary_by_met_slice.append(summary)
+            for vector in existing_vectors:
+                res = _evolve_vector(
```
tsankar, line 242: Another question that might be obvious to pycontrails devs, but why are there two evolution routines?

tsankar (commented):
Thank you for making this change so quickly!! My comments are mostly just clarifications, so feel free to ignore as you see fit.

tests/unit/test_cocip_grid.py
zebengberg: The model proceeds as normal from the start. Eventually we end up with a `GeoVectorDataset` whose time column exceeds the time dimension in `met`. When everything funnels down into the `RegularGridInterpolator`, the points will be out of bounds for the grid. With default parameters, we'll get nans. If `params["interpolation_bounds_error"] == True`, we'll get error. And if `params["interpolation_fill_value"] == None`, the value is extrapolated (we haven't played with this setting much).

In the default case, when we get a nan at the waypoint from interpolation, this nan will propagate forward into all the CoCiP calculations. Eventually there is a survival check, and the nan values will trigger an end of life condition. So the contrail waypoint will stop evolving.

pycontrails/models/cocipgrid/cocip_grid.py
zebengberg: Nope, it's definitely not obvious! Part if this is some lingering history that could be refactored to improve. Any new waypoints that haven't yet undergone any advection need to be initialized with the CoCiP downwash routine. Right now, this `_evolve_vector` function is responsible for both the downwash and the evolution. The two calls are for the new waypoints (for which the downwash has not yet been calculated), and for the partially-evolved waypoints. It might be a little cleaner to implement this different. I'll think about that.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -2069,7 +2220,7 @@ def _concat_verbose_dicts(
     # Concatenate the values and return
     ret: dict[str, np.ndarray] = {}
     for key in verbose_outputs_formation:
-        series_list = [v for d in verbose_dicts if (v := d.get(key)) is not None]
+        series_list = [v for d in verbose_dicts if d and (v := d.get(key)) is not None]
```
thabbott, line 2167: I didn't know about `:=`!

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -2202,8 +2354,8 @@ def _check_overlap(
     """
     if met_array.min() > grid_array.min() or met_array.max() < grid_array.max():
         warnings.warn(
-            f"Met data '{name}' does not overlap the grid domain along the {coord} axis. "
-            "This causes interpolated values to be nan, leading to meaningless results."
+            f"Met data '{name}' does not cover the source domain along the {coord} axis. "
```
thabbott, line 2301: I think "cover" is much better than "overlap"!

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -212,47 +200,105 @@ def eval(
...
-        summary = calc_intermediate_results(summary_by_met_slice)
-        return self._bundle_results(summary, verbose_dicts)
+        total_ef_summary = _aggregate_ef_summary(ef_summary)
+        return self._bundle_results(total_ef_summary, verbose_dicts)
+
+    def _maybe_downselect_met_rad(
+        self,
+        met: MetDataset | None,
+        rad: MetDataset | None,
+        time_end: np.datetime64,
```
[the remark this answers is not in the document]
zebengberg: I think the start_time is actually implicitly used here. There is now a `CocipGrid.timesteps` attribute that gets cooked up from the `source` time, the `max_age`, and the `dt_integration` param. The logic is opaque, but this function does typically slice `self.met` at both a beginning and end time.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -212,47 +200,105 @@ def eval(
...
+    ) -> tuple[MetDataset, MetDataset]:
+        """Downselect met and rad slices if necessary.
+
+        If ``self.params["downselect_met"]`` is True, :func:`_downselect_met` has
+        already performed a spatial downselection of the met data. This method
+        only downselects ``met`` and ``rad`` in the time domain.
```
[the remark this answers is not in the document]
zebengberg: ed08118d9e90d7739553e6e008fdd8aeff69813e

CHANGELOG.md
```diff
@@ -1,7 +1,15 @@
...
 # Changelog
 
-## v0.49.4 (unreleased)
+## v0.49.4
+
+### Breaking changes
+
+- Remove the `CocipGridParams.met_slice_dt` parameter. Now met downselection is handled automatically during contrail evolution. When the `met` and `rad` data passed into `CocipGrid` are not already loaded into memory, this update may make `CocipGrid` slightly more performant.
+- No longer explicitly load `met` and `rad` time slices into memory in `CocipGrid`. This now only occurs downstream when interpolation is performed. This change better aligns `CocipGrid` with other pycontrails models.
+- Remove the `cocipgrid.cocip_time_handling` module. Any useful tooling has been moved directly to the `cocipgrid.cocip_grid` module.
```
thabbott, line 10: My warnings 😭 😭 😭 

More seriously---unless I've missed something, the detailed warnings added to address #153 have been lost along with `cocip_time_handling.py`. I think we should keep these warnings, since they address a sharp edge and are referenced in the `CocipGrid` docstring. (Caveats about possible bias because I implemented those changes apply....) `_check_overlap` in `cocip_grid.py` seems like the right place to restore those warnings.

thabbott (approved):
Impressive work!

My comments are all pretty minor, but there are a couple (related to the warnings added to address #153 and to how time steps are determined) that I think are worth thinking about before merging.

pycontrails/models/cocipgrid/cocip_grid.py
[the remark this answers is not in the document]
thabbott: @zebengberg yes, I see that now. My confusion stemmed from the fact that downselection is triggered by looking at end times---but when you downselect, you slice starting from `idx - 1` and not `0`.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -342,6 +389,105 @@ def _bundle_results(
...
+        """Set the :attr:`timesteps` based on the ``source`` time range."""
+        source_time = self.source_time
+        tmin = source_time.min()
+        tmax = source_time.max()
+
+        tmin = pd.to_datetime(tmin)
+        tmax = pd.to_datetime(tmax)
+        dt = pd.to_timedelta(self.params["dt_integration"])
+
+        t_start = tmin.ceil(dt)
```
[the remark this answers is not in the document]
zebengberg: This is an interesting thought. Other folks using pycontrails have requested this sort of feature (it's especially useful for comparing CoCiP outputs to GOES imagery occurring with a fixed offset). What we have here is somewhat arbitrary ... I'm not sure what would make the most sense. I think someone could override or monkey patch `CocipGrid._set_timesteps` to customize the choice here. It might be cleaner to include an model parameter `timesteps_override` that could be a callable or numpy array.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -342,6 +389,105 @@ def _bundle_results(
...
+        t_end = tmax.floor(dt) + self.params["max_age"] + dt
+        self.timesteps = np.arange(t_start, t_end, dt)
+
+    def init_pbar(self) -> tqdm.tqdm | None:
```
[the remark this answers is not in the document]
zebengberg: Yes, that's a much simpler idea than what we have. Figuring out this a good way to track progress has been very annoying.

Two possible issues with using `timesteps` as the source of progress:

- `CocipGrid` doesn't necessarily make it through all timesteps. If no contrail waypoints survive at a given timestep, the model terminates.
- Most of the compute time would occur at `timestep = 0`.

Perhaps there is a hybrid approach we could take .... like some sort of discounting over timesteps.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -187,21 +185,11 @@ def eval(
             # in the source (we need to evolve contrails forward in time).
```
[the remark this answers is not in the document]
zebengberg: Thank you! It's slightly improved now. f0930bcfc0dbd10a2ba48e569fd419c2729b9cf0

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -212,47 +200,105 @@ def eval(
...
+        time_end: np.datetime64,
+    ) -> tuple[MetDataset, MetDataset]:
+        """Downselect met and rad slices if necessary.
+
+        If ``self.params["downselect_met"]`` is True, :func:`_downselect_met` has
+        already performed a spatial downselection of the met data. This method
+        only downselects ``met`` and ``rad`` in the time domain.
+        """
+        if met is None or time_end > met.variables["time"].values[-1]:
+            idx = np.searchsorted(self.met.variables["time"].values, time_end)
```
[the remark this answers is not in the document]
zebengberg: 1d8a1f25053895c857e5a915b05a196d3bfe0bb8

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -342,6 +389,105 @@ def _bundle_results(
     # Common Methods & Properties
     # ---------------------------
 
+    @property
```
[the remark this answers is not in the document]
zebengberg: I think we could set this as an instance attribute (we're inconsistent with this sort of thing all across pycontrails). I think I'll leave for now (this method is cheap), but we may want to do a cleanup along these lines later.

CHANGELOG.md
zebengberg: I've included this back in the `_set_timesteps` method. 143fdeab259d9a568a8e119778fa287d41c757b8

I've also tweaked one of my new tests to ensure this warning is emitted.

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -441,6 +441,8 @@ def _set_timesteps(self) -> None:
 
         t_start = tmin.ceil(dt)
         t_end = tmax.floor(dt) + self.params["max_age"] + dt
+        _check_met_rad_time(self.met, self.rad, tmin, t_end)
```
thabbott, line 444: For consistency, use `tmin` and `tmax` or `t_start` and `t_end`? Maybe you have a reason for doing this, though.

Otherwise, re-incorporated warnings look good!

thabbott (approved):

pycontrails/models/cocipgrid/cocip_grid.py
zebengberg: Good eye! I wanted to use the minimum and the maximum of the four variables `tmin`, `tmax`, `t_start`, `t_end`, which is why you see the inconsistency here. I'll leave a comment here.

zebengberg: 9f81078e61cf9a1df9976d4bbf354d4521555e2c

PR #157, opened 2024-02-28
Feature/unterstrasser 2016

roger-teoh:
Closes #XX

## Changes

Implement improved estimation of the survival fraction of contrail ice crystal number after the wake vortex phase (f_surv). The parameterised model is described by Unterstrasser (2016) and was developed using outputs provided by large eddy simulations. 

These changes replicates Fig. 4 of Karcher (2018), where f_surv now depends on the initial number of ice crystals. These effects are particularly important, especially in the 'soot-poor' scenario where the **number fraction** of contrail ice crystals that survives the wake vortex phase could be larger than the **mass fraction**, because the particles are larger in size. 

This also improves upon the existing assumption in CoCiP, where the survival fraction is estimated as the change in contrail ice water content (by mass) before and after the wake vortex phase. 

Unterstrasser, S., 2016. Properties of young contrails–a parametrisation based on large-eddy 
simulations. ACP, 16(4), pp.2059-2082. https://doi.org/10.5194/acp-16-2059-2016

Kärcher, B., 2018. Formation and radiative forcing of contrail cirrus. Nature communications, 9(1), p.1824.

## Tests

- [ ] CI tests pass

## Reviewer

@zebengberg


roger-teoh:
@zebengberg These changes need to be incorporated into the CoCiP grid function, which is not done yet. I look forward to getting your feedback to improve the computational efficiency of the code.

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+
+    Notes
+    -----
+    - See eq. (A6) and (A7) in :cite:`unterstrasserPropertiesYoungContrails2016`.
+    """
+    r_plume = 1.5 + 0.314 * wingspan
+    return 2.0 * 2.0 * np.pi * r_plume**2
+
+
+def z_desc_length_scale(
```
thabbott, line 647: This function isn't used anywhere. Can you add a test to make sure it's implemented correctly?

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -1,4 +1,24 @@
-"""Wave-vortex downwash functions."""
+"""Wave-vortex downwash functions.
+
+This module includes both equations from the original CoCiP model
+:cite:`schumannContrailCirrusPrediction2012` as well as improvements from
+:cite:`unterstrasserPropertiesYoungContrails2016`.
```
thabbott, line 5: Could you qualify in what sense the U16 version is an improvement? (Better agreement with LES?)

tests/unit/test_cocip.py
```diff
@@ -1596,3 +1604,60 @@ def test_cocip_met_nonuniform(
     assert np.all(contrail_time > t0)
     assert np.any(contrail_time < t1)
     assert np.any(contrail_time > t1)
+
```
thabbott, line 1607: Could you add two additional tests to check
- that `Cocip` runs with `improved_wake_vortex_ice_survival_fraction = True`? (No need to check values of outputs, just confirm that they're present.)
- that `CocipGrid` produces the expected error (https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) when run with `improved_wake_vortex_ice_survival_fraction = True`?

thabbott (commented):
This looks good! Thanks for your work on it. I have a few high-level comments, and left some specific ones inline.

1. I rebased this branch so changes from other branches don't show up in the diff. Can you double-check that none of your changes were lost?

2. Do we definitely want to describe the Unterstrasser update as an "improved" survival fraction? That feels like a pretty strong endorsement, and might make users wonder why it's not the default option. "Refined" is an alternative that (to me) acknowledges that it incorporates physical effects that are missing in the original CoCiP without necessarily suggesting that it's the best option in all scenarios. Happy to defer to you on this point.

3. I'd rather break functions for the Unterstrasser parameterization out into a separate file to make it easier to keep track of which functions are from the original CoCiP wake vortex model vs the Unterstrasser update. If you have time to do that, great! If not, let's get the new parameterization merged so you can use it and I'll can refactor later.

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+    # Solve non-linear equation numerically using the bisection method
+    # Did not use scipy functions because it is unstable when dealing with np.arrays
+    z_1 = np.zeros_like(rhi_issr)
+    z_2 = np.full_like(rhi_issr, 1000.0)
+    lhs = rhi_issr * thermo.e_sat_ice(air_temperature_issr) / air_temperature_issr
+
+    for _ in range(n_iter):
+        z_est = 0.5 * (z_1 + z_2)
+        rhs = (thermo.e_sat_ice(air_temperature_issr + 9.8e-3 * z_est)) / (
+            air_temperature_issr + 9.8e-3 * z_est
```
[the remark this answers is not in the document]
roger-teoh: I have just checked this. This is the dry adiabatic lapse rate, which is 9.8e-3 K per m. The 'T_lapse_rate = -0.065' in pycontrails.physics.constants is valid for the ICAO standard atmosphere from 0 km to 11 km. Should we include the dry adiabatic lapse rate in constants?

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+    # Did not use scipy functions because it is unstable when dealing with np.arrays
+    z_1 = np.zeros_like(rho_emit)
+    z_2 = np.full_like(rho_emit, 1000.0)
+
+    lhs = (thermo.e_sat_ice(air_temperature) / (constants.R_v * air_temperature)) + rho_emit
+
+    for _ in range(n_iter):
+        z_est = 0.5 * (z_1 + z_2)
+        rhs = thermo.e_sat_ice(air_temperature + 9.8e-3 * z_est) / (
+            constants.R_v * (air_temperature + 9.8e-3 * z_est)
```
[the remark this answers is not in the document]
roger-teoh: This is the dry adiabatic lapse rate, same as the comment above.

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+    - See eq. (3), (9), and (10) in :cite:`unterstrasserPropertiesYoungContrails2016`.
+    - For consistency in CoCiP, ``z_desc`` should be calculated using :func:`dz_max` instead of
+      using :func:`z_desc_length_scale`.
+    """
+    # Length scales
+    z_atm = z_atm_length_scale(air_temperature, rhi_0)
+    rho_emit = emitted_water_vapour_concentration(ei_h2o, wingspan, true_airspeed, fuel_flow)
+    z_emit = z_emit_length_scale(rho_emit, air_temperature)
+    z_total = z_total_length_scale(aei_n, z_atm, z_emit, z_desc)
+    return _ice_number_survival_fraction(z_total)
```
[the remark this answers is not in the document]
roger-teoh: Good suggestion!

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+    - See eq. (4) in :cite:`unterstrasserPropertiesYoungContrails2016`.
+    """
+    gamma_0 = _initial_wake_vortex_circulation(
+        wingspan, air_temperature, air_pressure, true_airspeed, aircraft_mass
+    )
+    n_bv = thermo.brunt_vaisala_frequency(air_pressure, air_temperature, dT_dz)
+    return ((8.0 * gamma_0) / (np.pi * n_bv)) ** 0.5
+
+
+def _initial_wake_vortex_circulation(
```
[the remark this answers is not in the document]
roger-teoh: I marked this function as non-public because it is currently not used elsewhere apart from the z_desc_length_scale` function. I don't have any strong preferences on this matter.

pycontrails/models/cocip/wake_vortex.py
```diff
@@ -373,3 +397,378 @@ def initial_contrail_depth(
...
+    -----
+    - This is a measure of the strength/intensity of the wake vortex circulation.
+    - See eq. (A1) in :cite:`unterstrasserPropertiesYoungContrails2016`.
+    """
+    b_0 = wake_vortex_separation(wingspan)
+    rho_air = thermo.rho_d(air_temperature, air_pressure)
+    return (constants.g * aircraft_mass) / (rho_air * b_0 * true_airspeed)
+
+
+def _ice_number_survival_fraction(z_total: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
```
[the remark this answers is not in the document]
roger-teoh: For this case, I have the preference to keep it as non-public, because it requires specific inputs, i.e., z_total, which can only be calculated from the upper level function `ice_particle_number_survival_fraction`

pycontrails/models/cocip/wake_vortex.py
roger-teoh: Yes, for now, this function is not used anywhere. It is in the Unterstrasser (2016) paper. However, to ensure consistency with the CoCiP wake vortex calculations, I have replaced the `z_desc` inputs with those calculated by CoCiP. I only included this function for completeness of the code, and it could be required in the future.

pycontrails/models/cocip/wake_vortex.py
roger-teoh: Parameterised ice crystal survival fraction based on large eddy simulations

roger-teoh:
Hi @thabbott , thank you for reviewing this branch and the detailed comments. I have included my answers to your high-level questions below:

_I rebased this branch so changes from other branches don't show up in the diff. Can you double-check that none of your changes were lost?_ 

I did a quick glance through and can confirm that none of the changes were lost. 

_Do we definitely want to describe the Unterstrasser update as an "improved" survival fraction? That feels like a pretty strong endorsement, and might make users wonder why it's not the default option. "Refined" is an alternative that (to me) acknowledges that it incorporates physical effects that are missing in the original CoCiP without necessarily suggesting that it's the best option in all scenarios. Happy to defer to you on this point._

Lets change it to "parameterised ice crystal survival fraction based on large eddy simulations". For now, I think it is best to not use it as the default option as it is still experimental. We need to run the global contrail simulations again to understand its effects.  

_I'd rather break functions for the Unterstrasser parameterization out into a separate file to make it easier to keep track of which functions are from the original CoCiP wake vortex model vs the Unterstrasser update. If you have time to do that, great! If not, let's get the new parameterization merged so you can use it and I'll can refactor later._

This is a very good suggestion. I had a think about it and it is probably a good idea not to include it in `cocip`, as it is not an original CoCiP implementation.

pycontrails/models/cocip/wake_vortex.py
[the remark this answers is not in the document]
thabbott: g/cp is the dry adiabatic lapse rate! You can derive it pretty easily from the first law of thermodynamics, ideal gas law, and hydrostatic balance: https://en.wikipedia.org/wiki/Lapse_rate#Dry_adiabatic_lapse_rate.

I'll add a variable called `dry_adiabatic_lapse_rate` and set it using values from `constants` so the physical meaning is clear in the code.

thabbott:
@roger-teoh: I've made changes to address the comments I left. Let me know if you see any issues!

One remaining issue to sort out: some of the unit tests are failing because changing the dry adiabatic lapse rate to `constants.g/constants.c_pd` (https://github.com/contrailcirrus/pycontrails/pull/157/commits/99c3b97f658d5f74d5bd3381462c78424f9d72b9#) slightly altered output from some of the Unterstrasser 2016 parameterizations. The relative changes are small---not more than 1 part in 100. If this doesn't bother you, I'll update expected values in the unit tests. If the changes will cause problems for you, we can just go back to hard-coded 9.8e-3.

roger-teoh:
@thabbott Thank you. I agree with the computation of the dry adiabatic lapse rate using g/c_pd. Yes, the minor changes are fine from my end. Feel free to update the unit tests.

pycontrails/models/cocip/unterstrasser_wake_vortex.py
```diff
@@ -0,0 +1,399 @@
...
+    -------
+    npt.NDArray[np.float64]
+        Fraction of ice particle number surviving the wake vortex phase
+    """
+    f_surv = 0.45 + (1.19 / np.pi) * np.arctan(-1.35 + (z_total / 100.0))
+    np.clip(f_surv, 0.0, 1.0, out=f_surv)
+    return f_surv
+
+
+def initial_contrail_depth_u2016(
```
mlshapiro, line 369: Given that this module is named for this paper, we don't need to suffix this method `u2016`

mlshapiro, line 369:
```suggestion
def initial_contrail_depth(
```

pycontrails/models/cocip/unterstrasser_wake_vortex.py
```diff
@@ -0,0 +1,399 @@
...
+    z_desc : npt.NDArray[np.float64]
+        Final vertical displacement of the wake vortex, ``dz_max`` in :mod:`wake_vortex.py`,
+        [:math:`m`].
+
+    Returns
+    -------
+    npt.NDArray[np.float64]
+        Fraction of contrail ice particle number that survive the wake vortex phase.
+
+    Notes
```
mlshapiro, line 70: Add `References` section even though this citation is already sprinked around
```
References
-----------
- :cite:`unterstrasserPropertiesYoungContrails2016`
```

pycontrails/models/cocip/unterstrasser_wake_vortex.py
mlshapiro: Is this method used by anything else? How does it compare with CoCip intial depth estimates?

tests/unit/test_unterstrasser_wake_vortex.py
```diff
@@ -0,0 +1,99 @@
+"""Test `unterstrasser_wake_vortex` module."""
+
+import numpy as np
+
+from pycontrails.models.cocip.unterstrasser_wake_vortex import (
+    _survival_fraction_from_length_scale,
+    initial_contrail_depth_u2016,
```
mlshapiro, line 7:
```suggestion
    initial_contrail_depth,
```

pycontrails/models/cocip/cocip.py
```diff
@@ -890,11 +891,27 @@ def _find_initial_persistent_contrails(self, met: MetDataset) -> None:
             air_temperature, air_pressure, air_pressure_1
         )
         iwc_1 = contrail_properties.iwc_post_wake_vortex(iwc, iwc_ad)
+
+        if self.params["improved_wake_vortex_ice_survival_fraction"]:
```
mlshapiro, line 895: Change the parameter to `"unterstrasser_ice_survival_fraction"`

mlshapiro, line 895:
```suggestion
        if self.params["unterstrasser_ice_survival_fraction"]:
```

mlshapiro (commented):
A few naming conventions

pycontrails/models/cocip/unterstrasser_wake_vortex.py
mlshapiro: Complete in 8607cb4e798dbab6eda57dea112c069213302574

PR #159, opened 2024-03-07
Feature/accf metrics

trdean1:
## Changes

> Adds now optional output that expresses forcing in both global averaged RF and ATR20

#### Internals

> Adds two new parameters to `CoCipParams`: `compute_atr20` (false by default) and `global_rf_to_atr20_factor` (0.0151 by default
> Global averaged RF is converted by taking EF (J) and dividing by the surface area of the earth (m^2) and the number of seconds in year, to give equivalented delta in global RF over a year needed to capture the same amount of energy. A constant multiple converts this to ATR20.
> `test_eval_persistent` sets `compute_atr20` and static test values are updated to include new outputs

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg 
> @thabbott


pycontrails/models/cocip/cocip.py
```diff
@@ -2372,6 +2383,12 @@ def calc_timestep_contrail_evolution(
     contrail_2["ef"] = energy_forcing_2
     contrail_2["age"][energy_forcing_2 == 0.0] = np.timedelta64(0, "ns")
 
+    if params["compute_atr20"]:
+        contrail_2["global_mean_rf"] = (
+            contrail_2["ef"] / constants.surface_area_earth / constants.seconds_per_year
+        )
+        contrail_2["atr20"] = params["global_rf_to_atr20_factor"] * contrail_2["global_mean_rf"]
```
thabbott, line 2390: It's not obvious to me why a year is the right length of time to use in the conversion from EF to RF. Is it just because the conversion factor (`params["global_rf_to_atr20_factor"]`) is defined for contrail RF averaged over a year?

If that's right, then I think we should change `"global_mean_rf"` to `"global_yearly_mean_rf"`, or something similar, to make it clearer that this is a quantity that depends on *both* the area and time period used for normalization.

pycontrails/models/cocip/cocip_params.py
```diff
@@ -136,6 +136,15 @@ class CocipParams(ModelParams):
...
 
+    #: Add additional metric of ATR20 and global mean RF to model output.
+    #: These are not standard CoCiP outputs but based on the derivation used
+    #: by :cite:`dietmullerPythonLibraryComputing2022`.
+    compute_atr20: bool = False
+
+    #: Constant factor used to convert global mean RF, [:math:`W m^{-2}`],
+    #: to ATR20, [:math:`K`], given by :cite:`dietmullerPythonLibraryComputing2022`.
+    global_rf_to_atr20_factor: float = 0.0151
+
```
thabbott, line 147: Similarly, for the sake of clarity I think it's worth changing "global mean RF" to something like "global- and year-mean contrail RF".

And, maybe it's worth adding a little bit more information about what ATR20 is? (At least a definition, e.g. "an estimate of the 20-year-average temperature change caused by the radiative forcing of a contrail", if that's accurate.)

thabbott (approved):
I left a couple of comments. My only substantial question is why you're computing RF averaged over a year (as opposed to some other length of time). I'm guessing that's connected to how the ATR20 conversion factor was derived, in which case I think we might want to call `"global_mean_rf"` something more specific.

Happy to discuss my comments if it's helpful, but otherwise feel free to merge after you've looked through them!

Oh, and @zebengberg will want you to update the changelog 😝. (It is actually quite helpful when we make new releases.)

pycontrails/models/cocip/cocip.py
trdean1: Yes, i should probably explicitly reference equation (S5) of the supplement to paper: https://gmd.copernicus.org/articles/16/3313/2023/gmd-16-3313-2023-discussion.html

PR #160, opened 2024-03-12
Bugfix/pytest removal error

thabbott:
## Changes

Bumps the minimum pytest version to 8.1 for consistency with the version currently being used in the release workflow.

Inconsistencies in pytest versions can produce unexpected errors in the release workflow. For example, https://github.com/contrailcirrus/pycontrails/actions/runs/8254181733 failed because the release workflow used pytest 8.1.1, and 8.1 is the first minor version without `pytest.PytestRemovedIn8Warning` (https://docs.pytest.org/en/7.1.x/backwards-compatibility.html?highlight=removedin). This wasn't caught in earlier testing (https://github.com/contrailcirrus/pycontrails/actions/runs/8253888054), which used pytest 7.4.4.

These changes are a bandaid solution in that they don't prevent similar future errors. These could occur if, for example, the release workflow starts using pytest >=9.1 but the test workflow stays on pytest <=9.0.

#### Internals

- Bump minimum pytest version to 8.1

## Tests

- [x] CI tests pass

## Reviewer

> @mlshapiro


mlshapiro:
This looks like a great solution, even if its a bandaid. I'm not sure we need to systematically fix these unless they are causing larger issues around.

PR #162, opened 2024-03-19
Feature/grid atr20

trdean1:
## Changes

> Adds optional ATR20 to CoCiPGrid model.

#### Internals

> Adds two parameters to `CoCipGridParams`, `compute_atr20` and `global_rf_to_atr20_factor`. Setting the former to `True` will add both `global_yearly_mean_rf` and `atr20` to the CoCiP output.

## Tests

- [x] CI tests pass

### DIFF
--- a/pycontrails/models/cocipgrid/cocip_grid_params.py
+++ b/pycontrails/models/cocipgrid/cocip_grid_params.py
@@ -1,4 +1,4 @@
-"""Default CocipGrid parameters."""
+"""Default :class:`CocipGrid` parameters."""
 
 from __future__ import annotations
 
@@ -17,7 +17,7 @@ class CocipGridParams(CocipParams):
     # Algorithm
     # ---------
 
-    #: Approximate size of a typical `np.array` used with in CoCiP calculations.
+    #: Approximate size of a typical :class:`numpy.ndarray` used with in CoCiP calculations.
     #: The 4-dimensional array defining the waypoint is raveled and split into
     #: batches of this size.
     #: A smaller number for this parameter will reduce memory footprint at the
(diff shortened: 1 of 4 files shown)


pycontrails/models/cocipgrid/cocip_grid_params.py
```diff
@@ -136,3 +136,13 @@ class CocipGridParams(CocipParams):
...
+
+    #: Add additional metric of ATR20 and global yearly mean RF to model output.
+    #: These are not standard CoCiP outputs but based on the derivation used
+    #: in the first supplement to :cite:`yin_predicting_2023`. ATR20 is defined
+    #: as the average temperature response over a 20 year horizon.
+    compute_atr20: bool = False
+
+    #: Constant factor used to convert global- and year-mean RF, [:math:`W m^{-2}`],
+    #: to ATR20, [:math:`K`], given by :cite:`yin_predicting_2023`.
+    global_rf_to_atr20_factor: float = 0.0151
```
thabbott, line 148: Are these additions necessary? I think both parameters should be inherited from `CocipParams`.

tests/unit/test_cocip_grid.py
```diff
@@ -288,6 +288,41 @@ def grid_results(
...
+
+    # Pin a few values
+    rel = 1e-2
+    assert ds["atr20_per_m"].mean() == pytest.approx(2.504e-18, rel=rel)
+    assert ds["global_yearly_mean_rf_per_m"].mean() == pytest.approx(1.659e-16, rel=rel)
+
+
 ##############################################################
 # NOTE: No tests below here will run unless BADA is available.
 ##############################################################
```
thabbott, line 328: Could you move this comment block above the previous two functions so it marks the place where tests will start to be skipped?

thabbott (commented):
Looks good! Just a couple of minor comments---ok by me if you merge once you've taken a look at them.

PR #165, opened 2024-03-20
Fix/short flight resampling

thabbott:
Addresses https://github.com/contrailcirrus/flights-pipeline/issues/56

## Changes

Updates flight resampling logic to ensure that resampled waypoints include any and all times between flight start and end times that are a multiple of the resampling frequency.

#### Breaking changes

Updates to flight resampling logic may add an additional waypoint to some flights after resampling, and may result in `Flight.resample_and_fill` returning a flight with a single waypoint rather than an empty flight.

#### Fixes

Flight resampling aligns with expected behavior for very short flights, which is now detailed in the `Flight.resample_and_fill` docstring.

## Tests

- [x] CI tests pass

## Reviewer

@nickmasson

### DIFF
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -2063,11 +2063,9 @@ def _resample_to_freq(df: pd.DataFrame, freq: str) -> tuple[pd.DataFrame, pd.Dat
 
     # Manually create a new index that includes all the original index values
     # and the resampled-to-freq index values.
-    t0 = df.index[0]
+    t0 = df.index[0].ceil(freq)
     t1 = df.index[-1]
-    t = pd.date_range(t0, t1, freq=freq, name="time").floor(freq)
-    if t[0] < t0:
-        t = t[1:]
+    t = pd.date_range(t0, t1, freq=freq, name="time")
 
     concat_arr = np.concatenate([df.index, t])
     concat_arr = np.unique(concat_arr)
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -787,6 +787,13 @@ class Flight(GeoVectorDataset):
         Waypoints are resampled according to the frequency ``freq``. Values for :attr:`data`
         columns ``longitude``, ``latitude``, and ``altitude`` are interpolated.
 
+        Resampled waypoints will include all multiples of ``freq`` between the flight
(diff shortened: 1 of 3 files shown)


tests/unit/test_flight.py
```diff
@@ -1108,6 +1111,19 @@ def test_resample_and_fill_short_flight() -> None:
     fl3 = fl.resample_and_fill("1 min", keep_original_index=True)
     assert len(fl3) == 2
 
+    fl = Flight(
```
mlshapiro, line 1114: Can you add an example of the other case, e.g.

```
...
time=[np.datetime64("2022-01-01T00:01:03"), np.datetime64("2022-01-01T00:01:59")],
```

and assert that the length is 0?

pycontrails/core/flight.py
```diff
@@ -2063,11 +2070,9 @@ def _resample_to_freq(df: pd.DataFrame, freq: str) -> tuple[pd.DataFrame, pd.Dat
 
     # Manually create a new index that includes all the original index values
     # and the resampled-to-freq index values.
-    t0 = df.index[0]
+    t0 = df.index[0].ceil(freq)
```
mlshapiro, line 2073: Clever logic!

mlshapiro (approved):

nickmasson (commented):
thanks @thabbott !

this appears to be behaving as expected from an external consumer's perspective...

I pulled the pycontrails version from this feature branch into the [flights-pipeline](https://github.com/contrailcirrus/flights-pipeline/tree/triage/test-pycontrails-fix).

Given [these two test cases](https://github.com/contrailcirrus/flights-pipeline/blob/9f77666bd3b52e8d1da9be8f80e3af6a6bc3527e/spire-ingest-resample-worker/tests/stubs/stub.py#L4).
Both are less than a minute interval's span (first waypoint to last). The inter minute spans (crossing a minute timestamp), now returns a resampled set of length 1.
The intra minute will still return resampled of length 0.

I don't have any tests in place to confirm the actual value(s) returned by interpolation, but, in as far as expected data-mongering, this looks great 🚀

tests/unit/test_flight.py
thabbott: There's an example of the other case earlier in this test (not visible in the diff because it was already there!)

PR #179, opened 2024-04-19
Feature/model level datalibs

### LINKED ISSUE #173: Model-level ECMWF datalibs
**Description**

We provide datalibs for accessing pressure-level [ERA5](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/ecmwf/era5.py) and [HRES](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/ecmwf/hres.py) data. We should implement similar interfaces to model-level ERA5 and HRES data (likely in two separate datalibs).

Recommended approach:

- [ ] Retrieve model-level GRIB files by submitting MARS requests to the [Copernicus Climate Data Store](https://cds.climate.copernicus.eu/#!/home) via the [CDS API](https://cds.climate.copernicus.eu/api-how-to) for ERA5 and to the [ECMWF operational archive](https://apps.ecmwf.int/archive-catalogue/?class=od) via the [ECMWF API client](https://pypi.org/project/ecmwf-api-client/) for HRES.
- [ ] Consider caching retrieved GRIB files using a `DiskCacheStore`, particularly if this allows [more efficient MARS requests](https://confluence.ecmwf.int/display/UDOC/Guidelines+to+write+efficient+MARS+requests).
- [ ] Retrieve model-level data at the target horizontal resolution, and interpolate to target pressure levels using [`metview`](https://metview.readthedocs.io/en/latest/).
- [ ] Consider including conveniences (parallel workers, cleanup of `metview` temporary files) similar to the [ARCO ERA5 datalib](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/ecmwf/arco_era5.py) provides access to model-level ERA5 data, but only through mid-2023 and not for ensemble members).
- [ ] For ERA5, include interfaces to nominal analyses ([example](https://apps.ecmwf.int/data-catalogues/era5/?stream=oper&levtype=ml&expver=1&month=jan&year=2019&type=an&class=ea)) and ensemble members ([example](https://apps.ecmwf.int/data-catalogues/era5/?stream=enda&levtype=ml&expver=1&month=jan&year=2019&date=2019-01-01&type=an&class=ea)).
- [ ] For HRES, implement an interface to nominal forecasts ([example](https://apps.ecmwf.int/archive-catalogue/?stream=oper&levtype=ml&time=00%3A00%3A00&expver=1&month=jan&year=2019&date=2019-01-01&type=fc&class=od)), and confirm that individual ensemble members are not available in the operational archive.

**Alternatives**

- The [ARCO ERA5 datalib](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/ecmwf/arco_era5.py) provides access to model-level ERA5 data, but only through mid-2023 and not for ensemble members.

thabbott:
Closes #173 

## Changes

Implements interfaces for model-level ERA5 and HRES data.

### Features

- Add `ModelLevelERA5` and `ModelLevelHRES` interfaces for accessing ERA5 and HRES data on model levels.
- Update [ECMWF tutorial notebook](https://py.contrails.org/notebooks/ECMWF.html) with instructions for using model-level datalibs.

### Internals

- Create static copy of dataframe for determining pressure at ECMWF model levels.
- Extract model-level utilities in `ARCOERA5` to their own module for reuse in `ModelLevelERA5` and `ModelLevelHRES`.
- Update Makefile so `make doctest` and `make nb-test` use data cached by `make ensure-era5-cached` when run locally.

## Tests

- [x] CI tests pass

## Reviewer

> @mlshapiro
> @trdean1


trdean1 (approved):
Overall looks great!

It might be good to put some tests into `tests/unit/test_ecmwf.py` that at least test some of the logic in the `__init__` methods

thabbott:
Yes, testing those parts of the new datalibs would be good!

I've added tests to `test/unit/test_ecmwf` that cover portions of `ModelLevelERA5` and `ModelLevelHRES` that don't use metview and don't access upstream APIs. I've parameterized tests by datalib where it makes sense, so these changes should also expand test coverage for other ECMWF datalibs.

Makefile
```diff
@@ -165,7 +165,7 @@ cache-era5-gcp: ensure-era5-cached
 	gcloud storage cp -r -n .doc-test-cache/* gs://contrails-301217-unit-test/doc-test-cache/
 
 doctest: ensure-era5-cached ensure-gcp-credentials
-	pytest --doctest-modules \
+	PYCONTRAILS_CACHE_DIR="$(CURDIR)/.doc-test-cache" pytest --doctest-modules \
```
mlshapiro, line 168: These ENV variables are set in the Github action (See for example [doctest action](https://github.com/contrailcirrus/pycontrails/blob/2c06549b985ee8af7158163ca908fd7af7a4997d/.github/workflows/doctest.yaml#L22)

I assume this will override the global ENV, but perhaps this is generic as to work everywhere? Just make sure you have that reconciled.

pycontrails/datalib/ecmwf/__init__.py
```diff
@@ -6,6 +6,8 @@
 from pycontrails.datalib.ecmwf.era5 import ERA5
 from pycontrails.datalib.ecmwf.hres import HRES
 from pycontrails.datalib.ecmwf.ifs import IFS
+from pycontrails.datalib.ecmwf.model_level_era5 import ModelLevelERA5
```
mlshapiro, line 9: is there a reason to have these outside the `.era5`, `.hres` modules? 

I could see `datalib.ecmwf.era5` containing any ERA5 related datalib, or I do also see the credence in having one class per module, etc.

Not too opinionated here, just want to make sure the org feels logical.

pycontrails/datalib/ecmwf/static/model_level_dataframe_v20240418.csv
mlshapiro, line 1: NIce - I imagine this won't ever change? But it could?

pycontrails/datalib/ecmwf/model_level_era5.py
```diff
@@ -0,0 +1,501 @@
...
+
+import pycontrails
+from pycontrails.core import cache, datalib, met_var
+from pycontrails.core.met import MetDataset, MetVariable
+from pycontrails.datalib.ecmwf import variables as ecmwf_var
+from pycontrails.datalib.ecmwf.common import ECMWFAPI
+from pycontrails.datalib.ecmwf.model_levels import pressure_levels_at_model_levels
+from pycontrails.utils import dependencies, temp
+
+MODEL_LEVEL_VARIABLES = [
```
mlshapiro, line 50: Is there a reason these are different than the pressure level variables? I would expect thes variables to be the same as in the other `ERA5` class

pycontrails/datalib/ecmwf/model_level_era5.py
```diff
@@ -0,0 +1,501 @@
...
+    ecmwf_var.Divergence,
+    ecmwf_var.CloudAreaFractionInLayer,
+    ecmwf_var.SpecificCloudIceWaterContent,
+    ecmwf_var.SpecificCloudLiquidWaterContent,
+]
+
+ALL_ENSEMBLE_MEMBERS = list(range(10))
+
+
+class ModelLevelERA5(ECMWFAPI):
```
mlshapiro, line 66: Annoying naming convention - would it make more sense to call this `ERA5ModelLevel`

pycontrails/datalib/ecmwf/model_level_era5.py
```diff
@@ -0,0 +1,501 @@
...
+            )
+
+        try:
+            self.cds = cdsapi.Client(url=self.url, key=self.key)
+        # cdsapi throws base-level Exception
+        except Exception as err:
+            raise CDSCredentialsNotFound from err
+
+
+def _download_convert_cache_handler(
```
mlshapiro, line 403: Is there a reason this is outside the class? Is it re-used elsewhere?

pycontrails/datalib/ecmwf/model_level_era5.py
```diff
@@ -0,0 +1,501 @@
...
+
+        Returns
+        -------
+        dict[str, str]:
+            MARS request for submission to Copernicus CDS.
+        """
+        unique_dates = set(t.strftime("%Y-%m-%d") for t in times)
+        unique_times = set(t.strftime("%H:%M:%S") for t in times)
+        # param 152 = log surface pressure, needed for metview level conversion
+        grib_params = set(self.variable_ecmwfids + [152])
```
mlshapiro, line 364: Perfect - just confirming you will 152 is the `lnsp` needed for conversion - looks like you already took this into account!

pycontrails/datalib/ecmwf/model_level_era5.py
```diff
@@ -0,0 +1,501 @@
...
+                    fs_pl = mv.merge(fs_pl, f_pl)
+
+            # Create, validate, and cache dataset
+            ds = fs_pl.to_dataset()
+            ds = ds.rename(isobaricInhPa="level").expand_dims("time")
+            ds.attrs["pycontrails_version"] = pycontrails.__version__
+            era5.cache_dataset(ds)
+
+
+class CDSCredentialsNotFound(Exception):
```
mlshapiro, line 500: Could this just be imported from the other `era5` module?

pycontrails/datalib/ecmwf/model_level_hres.py
```diff
@@ -0,0 +1,511 @@
...
+import pycontrails
+from pycontrails.core import cache, datalib, met_var
+from pycontrails.core.met import MetDataset, MetVariable
+from pycontrails.datalib.ecmwf import variables as ecmwf_var
+from pycontrails.datalib.ecmwf.common import ECMWFAPI
+from pycontrails.datalib.ecmwf.model_levels import pressure_levels_at_model_levels
+from pycontrails.utils import dependencies, temp
+from pycontrails.utils.types import DatetimeLike
+
+MODEL_LEVEL_VARIABLES = [
```
mlshapiro, line 39: Are these unique to Model Levels?

pycontrails/datalib/ecmwf/model_level_hres.py
```diff
@@ -0,0 +1,511 @@
...
+                name="ModelLevelHRES._set_server method",
+                package_name="ecmwf-api-client",
+                module_not_found_error=e,
+                pycontrails_optional_package="ecmwf",
+            )
+
+        self.server = ECMWFService("mars", url=self.url, key=self.key, email=self.email)
+
+
+def _download_convert_cache_handler(
```
mlshapiro, line 424: Similar to comment to era5 - is there a reason this is outside the class?

pycontrails/datalib/ecmwf/model_level_hres.py
```diff
@@ -0,0 +1,511 @@
...
+        (:py:class:`datetime.datetime`, :class:`pandas.Timestamp`, :class:`numpy.datetime64`)
+        specifying the (start, end) of the date range, inclusive.
+        All times will be downloaded in a single GRIB file, which
+        ensures that exactly one request is submitted per file on tape accessed.
+        If ``forecast_time`` is unspecified, the forecast time will
+        be assumed to be the nearest synoptic hour available in the operational archive (00 or 12).
+        All subsequent times will be downloaded for relative to :attr:`forecast_time`.
+    variables : datalib.VariableInput
+        Variable name (i.e. "t", "air_temperature", ["air_temperature, specific_humidity"])
+    pressure_levels : datalib.PressureLevelInput, optional
```
mlshapiro, line 98: Is it worth keeping the `pressure levels` input to these classes?  Would it work if I passed in arbitrary pressure levels that were a different length than the `levels` input?

mlshapiro (approved):
Amazing work - I don't have much to comment here, mostly naming conventions and trying to not open up too many hairy edges with inputs.

Just want to confirm that this will cache the pressure level data (not the original Grib or model level data)

thabbott:
> Amazing work - I don't have much to comment here, mostly naming conventions and trying to not open up too many hairy edges with inputs.
> 
> Just want to confirm that this will cache the pressure level data (not the original Grib or model level data)

This will always cache the pressure-level data, and will optionally cache the original grib data if `cache_grib=True` is passed as a keyword argument. (I think you'd only want to do that if you're changing something in the grib processing and don't want to wait for the grib file to be re-downloaded each time you process it, so I think of this option as an aid for developers rather than something useful for users.)

pycontrails/datalib/ecmwf/__init__.py
thabbott: I don't have a strong opinion, either--I made this choice mostly for consistency with the ARCO ERA5 datalib, which is in its own file.

pycontrails/datalib/ecmwf/static/model_level_dataframe_v20240418.csv
thabbott: We'd need to update it if something changes about the vertical discretization in ECMWF models (e.g., if they add more vertical levels).

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: Yes! It's because not all ERA5 pressure-level variables are available on model levels. (Model levels aren't missing anything we care about, just RH and PV).

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: I don't have a strong opinion--happy to change to `ERA5ModelLevel` and `HRESModelLevel` if you prefer!

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: Nope. I think I was wondering initially if I might be able to re-use this function for the model-level HRES datalib, but they're different enough for that not to make sense. I'll move it inside the class!

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: Yes... will fix that

pycontrails/datalib/ecmwf/model_level_hres.py
thabbott: Nope--will move this inside too!

pycontrails/datalib/ecmwf/model_level_hres.py
thabbott: Yes! `pressure_levels` can be anything you want, with the caveat that you might get some NaNs if you set the `levels` input to something that doesn't download enough model levels to cover the range of pressure levels you request.

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: 2a4776b3b0a010206afe122927a3b399f4266f2d

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: ff29b019df9b83a4d08354566e3842d99c2ed3a4

pycontrails/datalib/ecmwf/model_level_hres.py
thabbott: 0c24a63e661e903787c7b0c6b94bac1e9a667e19

pycontrails/datalib/ecmwf/model_level_era5.py
thabbott: c4a63967c848ec05fe1c08990ed4c23cf05dba78

PR #181, opened 2024-04-29
Histogram matching for model-level ERA5 data

thabbott:
#### Features

- Add `HistogramMatching` humidity scaling calibrated for model-level ERA5 data.

Creating an instance of `HistogramMatching` without explicitly selecting the model- or pressure-level calibration will use the pressure-level calibration by default (which preserves current behavior) but warn users that a future release will default to the model-level calibration.

## Tests

- [x] CI tests pass

## Reviewer

> @trdean1 
> @mlshapiro


mlshapiro (approved):
This all looks good to me! Glad you could re-use the existing infrastructure

trdean1:
Looks great to me as well!

PR #183, opened 2024-05-07
Fix/geodesic interpolation

### LINKED ISSUE #182: Geodesic interpolation with missing altitude data
**Description**

`Flight.resample_and_fill` uses geodesic interpolation when the distance between waypoints, as measured by `Flight.segment_length`, exceeds a threshold. However, `Flight.segment_length` returns NaN for segments bounded by waypoints with NaN altitudes. As a result, the presence of NaN altitudes can prevent geodesic interpolation from being used to fill gaps between waypoints separated by large horizontal distances.

**Details**

- Version: 0.50.2
- Module: core.flight
- OS: ubuntu

**Steps to Reproduce**

```python
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import numpy as np

from pycontrails.core import Flight

longitude = [-120, -70, 1]
latitude = [45, 45, 45]
time = ["2024-01-01 00:00", "2024-01-01 05:00", "2024-01-01 12:00"]

flight = Flight(
    longitude = longitude,
    latitude = latitude,
    altitude_ft = [35_000, 35_000, 35_000],
    time=time
).resample_and_fill("10min")

plt.subplot(111, projection=ccrs.PlateCarree())
plt.plot(flight["longitude"], flight["latitude"], 
    "k-", transform=ccrs.PlateCarree(), label="expected")

flight = Flight(
    longitude = longitude,
    latitude = latitude,
    altitude_ft = [35_000, np.nan, 35_000],
    time=time
).resample_and_fill("10min")

plt.plot(flight["longitude"], flight["latitude"], 
    "r-", transform=ccrs.PlateCarree(), label="missing altitude")

plt.legend()
plt.gca().set_extent([-140, 20, 0, 90])
plt.gca().coastlines()
plt.show()
```

image

thabbott:
Closes #182

## Changes

This PR changes the condition for using geodesic interpolation in `Flight.resample_and_fill` from one based on total segment length to one based on (horizontal) great circle distance. This seems reasonable on physical grounds (the error in using linear interpolation to estimate the position of a flight following a great circle is a function of the horizontal distance between waypoints), and fixes the bug documented in #182:

image

#### Breaking changes

- Geodesic interpolation is now used in `Flight.resample_and_fill` when the great circle distance between waypoints (rather than the total segment length including vertical displacement) exceeds a threshold. This may change the interpolation method used when resampling flight segments with lengths close to the geodesic interpolation threshold.

#### Fixes

- Use horizontal great circle distance to determine whether geodesic interpolation is used in `Flight.resample_and_fill`. This ensures geodesic interpolation is used between sufficiently distant waypoints even when one or both waypoints contain NaN altitude data.

## Tests

- [x] CI tests pass

## Reviewer

> @mlshapiro

### DIFF
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -454,7 +454,7 @@ class Flight(GeoVectorDataset):
         `np.nan` appended so the length of the output is the same as number of waypoints.
 
         To account for vertical displacements when computing segment lengths,
-        use :meth:`segment_length`
+        use :meth:`segment_length`.
 
         Returns
         -------
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -447,6 +447,48 @@ class Flight(GeoVectorDataset):
 
         return segment_duration(self.data["time"], dtype=dtype)
 
+    def segment_haversine(self) -> npt.NDArray[np.float64]:
+        """Compute Haversine (great circle) distance between flight waypoints.
+
+        Helper function used in :meth:`resample_and_fill`.
+        `np.nan` appended so the length of the output is the same as number of waypoints.
+
+        To account for vertical displacements when computing segment lengths,
+        use :meth:`segment_length`
+
+        Returns
+        -------
+        npt.NDArray[np.float64]
+            Array of great circle distances in [:math:`m`] between waypoints
+
+        Raises
+        ------
+        NotImplementedError
+            Raises when attr:`attrs["crs"]` is not EPSG:4326
+
+        Examples
+        --------
+        >>> from pycontrails import Flight
+        >>> fl = Flight(
+        ... longitude=np.array([1, 2, 3, 5, 8]),
+        ... latitude=np.arange(5),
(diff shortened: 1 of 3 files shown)


pycontrails/core/flight.py
```diff
@@ -447,6 +447,48 @@ def segment_duration(self, dtype: npt.DTypeLike = np.float32) -> npt.NDArray[np.
...
         return segment_duration(self.data["time"], dtype=dtype)
 
+    def segment_haversine(self) -> npt.NDArray[np.float64]:
+        """Compute Haversine (great circle) distance between flight waypoints.
+
+        Helper function used in :meth:`resample_and_fill`.
+        `np.nan` appended so the length of the output is the same as number of waypoints.
+
+        To account for vertical displacements when computing segment lengths,
+        use :meth:`segment_length`
```
mlshapiro, line 457:
```suggestion
        use :meth:`segment_length`.
```

mlshapiro (approved):
@thabbott this looks great to me. Thanks for includeing the doctests and nice docstring!

PR #192, opened 2024-05-20
Feature/upper bound polygons

### LINKED ISSUE #191: polygon generator | support for cooling (negative ef) polygons
**Description**
At present, the geojson polygon generator will only generate polygons (with expected behavior) for positive thresholds.

This Issue proposes adding support for cooling (negative ef) polygons.

**Suggested Behavior**
For thresholds greater (or equal to) zero ([J/m]), then business as usual.

For thresholds less than zero, then polygon represent all region with J/m less than the defined threshold.

[Reference.](https://github.com/contrailcirrus/pycontrails/blob/3816d491fb56debbf12b36211b26f4e521d93d10/pycontrails/core/polygon.py#L342)

thabbott:
Closes #191

## Changes

Adds a `lower_bound` keyword argument to `polygon.find_multipolygon` that controls whether `threshold` is used as a lower or upper bound on the value of regions inside polygons. The default value (`lower_bound=True`) preserves existing behavior (using `threshold` as a lower bound). 

To find cooling polygons where `threshold` is an upper bound on EF, pass `lower_bound=False`:
```python
threshold = -1e7
mp = polygon.find_multipolygon(ef, threshold, lower_bound=False, ...)
```

#### Features

Add ability to use `polygon.find_multipolygon` to find regions with values below a threshold.

## Tests

- [x] CI tests pass

## Reviewer

> @nickmasson, @mlshapiro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,14 @@
 
 - Geodesic interpolation is now used in `Flight.resample_and_fill` when the great circle distance between waypoints (rather than the total segment length including vertical displacement) exceeds a threshold. This may change the interpolation method used when resampling flight segments with lengths close to the geodesic interpolation threshold.
 - Fixed typo in `thermo.c_pm`  will decrease computed values of moist heat capacity with non-zero specific humidity. We expect the downstream impact on contrail predictions by `ISSR`, `SAC`, `PCR, and `Cocip` models to be minimal.
+- `np.nan` is now used as the default `fill_value` in `MetDataArray.to_polygon_feature` and `MetDataArray.to_polygon_feature_collection`. This ensures that NaN values are never included in polygon interiors unless a non-NaN `fill_value` is explicitly passed as a keyword argument.
 
 ### Features
 
 - Add `ERA5ModelLevel` and `HRESModelLevel` interfaces for accessing ERA5 and HRES data on model levels.
 - Update [ECMWF tutorial notebook](https://py.contrails.org/notebooks/ECMWF.html) with instructions for using model-level datalibs.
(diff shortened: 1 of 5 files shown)


nickmasson (commented):
I live this design, as it is explicit behavior and requires the caller to mind the intended behavior.

To make this drop-in usable, however, we should bubble up the lower_bound param to other methods that use this method in composing their behavior.

For instance, [this is the pycontrails method](https://github.com/contrailcirrus/pycontrails/blob/3816d491fb56debbf12b36211b26f4e521d93d10/pycontrails/core/met.py#L1792) that we are using in the api-preprocessor.
There may be other places as well.

Alternatively, we could have it be implicit behavior, and infer the directionality based on the sign of the threshold (I'm not insisting we do this necessarily, the tradeoffs between explicit behavior and code complexity seem balanced).

--

Separate question. How is a zero value within the threshold range expected to behave?
If a contrail is not formed, will it be a nan value in the grid, or will it be zeroed?
This seems like a relevant consideration as we extend the behavior of the polygon generator.  As a user, it isn't immediately clear to me if including zero in the range means I'm including all regions where contrails are not formed, or only areas where contrails are formed, but the value of the radiative forcing is within my range.

thabbott:
The most recent changes surface the `lower_bound` keyword in `MetDataArray.to_polygon_feature` and `MetDataArray.to_polygon_feature_collection`. For completeness, I made a similar change in `MetDataArray.to_polyhedra`. @nickmasson, do these changes give you the tooling you need to generate polygons with negative EF?

mlshapiro (approved):
Looks great, thanks @thabbott. Thanks for including changes and well documenting the additions.

PR #195, opened 2024-05-27
Feature/ps-update-20240524

roger-teoh:
Closes #XX

## Changes

Update Poll-Schumann aircraft performance model to version-20240524

#### Features
- PS model: Support four aircraft types, including `E75L`, `E75S`, `E290`, and `E295` 
- PS model: Integrate `ps_synonym_list` to increase PS model aircraft type coverage to 102
- PS model: Account for increase in fuel consumption due to engine deterioration between maintenance cycle

#### Internals

- Calculate rate of climb and descent (ROCD) using hydrostatic equation to improve accuracy
- Update unit tests
- PS model: Move engine performance buffer from `c_t_available` to `tr_max`
- Set`tr_max` buffer to +20%, as recommended by Ian Poll

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer


roger-teoh:
@trdean1 One thing to pay more attention during the review of this branch is the change in calculating `segment_rocd` in the flight class. If `air_temperature` is available, then the segment_rocd  estimate will be more accurate with a temperature correction (More important when used in aircraft performance models, probably not required when cleaning the Spire raw ADS-B data). If it is not available, then segment_rocd can still be used too. This might affect usage in other areas.

pycontrails/core/flight.py
```diff
@@ -2165,12 +2181,22 @@ def segment_rocd(
...
     npt.NDArray[np.float64]
         Rate of climb and descent over segment, [:math:`ft min^{-1}`]
 
+    Notes
+    -----
+    The hydrostatic equation will be used to estimate the ROCD if `air_temperature` is provided.
+    This will improve the accuracy of the estimated ROCD with a temperature correction. The
+    estimated ROCD with the temperature correction are expected to differ by up to +-5% compared to
+    those without the correction. These differences are important when the ROCD estimates are used
+    as inputs to aircraft performance models.
```
trdean1, line 2198: Do you have any references that could be cited as to where this correction came from?

trdean1 (approved):
This looks great to me!  If you have a source to cite as to where the hydrostatic correction came from that would be useful to add, but it is also okay if you don't.  I played around with the outputs a bit and they all seem reasonable ot me.

PR #200, opened 2024-05-30
Landsat and Sentinel-2 datalibs

thabbott:
### Features

- Add `Landsat` and `Sentinel` datalibs for searching, retrieving, and visualizing Landsat 8-9 and Sentinel-2 imagery. The datalibs include:
    - Tools for querying Landsat and Sentinel-2 imagery for intersections with user-defined regions (`landsat.query`, `sentinel.query`) or flights (`landsat.intersect`, `sentinel.intersect`). These tools use BigQuery tables and require a Google Cloud Platform account with access to the BigQuery API.
    - Tools for downloading and visualizing imagery from Landsat (`Landsat`) and Sentinel-2 (`Sentinel`). These tools retrieve data anonymously from Google Cloud Platform storage buckets and can be used without a Google Cloud Platform account.
- Add tutorial notebooks demonstrating how to use `Landsat` and `Sentinel` datalibs to find flights in high-resolution satellite imagery.

### Internals

- Add unit tests for Landsat and Sentinel search tools and datalibs, but disable any tests that retrieve imagery data when running tests in GitHub Actions to limit resource consumption in GitHub runners. Users can disable these unit tests locally by setting the environment variable `GITHUB_ACTIONS=true`.

## Tests

- [x] CI tests pass

## Reviewer

> @mlshapiro


pycontrails/datalib/_leo_utils/search.py
```diff
@@ -0,0 +1,259 @@
...
+            decoded = geojson.Feature(geometry=geojson.loads(self.extent))
+        except Exception as exc:
+            msg = "extent cannot be converted to GeoJSON structure"
+            raise ValueError(msg) from exc
+        if not decoded.is_valid:
+            msg = "extent is not valid GeoJSON"
+            raise ValueError(msg)
+
+
+def track_to_geojson(lon: np.ndarray, lat: np.ndarray) -> str:
```
mlshapiro, line 61: Should we reconcile this with https://github.com/contrailcirrus/pycontrails/blob/b08e971f87166710eb41f4b61ba86eab687056f9/pycontrails/core/flight.py#L1351 and https://github.com/contrailcirrus/pycontrails/blob/b08e971f87166710eb41f4b61ba86eab687056f9/pycontrails/core/flight.py#L1376

(Also for reference vector has one too https://github.com/contrailcirrus/pycontrails/blob/b08e971f87166710eb41f4b61ba86eab687056f9/pycontrails/core/vector.py#L1955


Can we abstract those methods in a way that works for this method? or at least use the same approaches in both? Then we can add a "See Also" section to the docstring with 

```
See Also
---------
:meth:`Flight.to_geojson_linestring`
:meth:`Flight.to_geojson_multilinestring`

(We ended up writing our own implementation to avoid the `geojson` dependency - although that dependency looks OK to include even at the core level.)

File: pycontrails/datalib/sentinel.py [hunk 2]
```diff
@@ -0,0 +1,506 @@
...
+        By default, returns imagery base URL, granule ID, and sensing time.
+
+    Returns
+    -------
+    pd.DataFrame
+        Query results in pandas DataFrame
+
+    See Also
+    --------
+    `func`:search.query:
```
mlshapiro, line 102:
```suggestion
    :func:`search.query`
```

pycontrails/datalib/sentinel.py
```diff
@@ -0,0 +1,506 @@
...
+        By default, returns imagery base URL, granule ID, and sensing time.
+
+    Returns
+    -------
+    pd.DataFrame
+        Query results in pandas DataFrame
+
+    See Also
+    --------
+    :func"`search.intersect`
```
mlshapiro, line 139:
```suggestion
    :func:`search.intersect`
```

pycontrails/datalib/landsat.py
```diff
@@ -0,0 +1,562 @@
...
+        By default, returns imagery base URL and sensing time.
+
+    Returns
+    -------
+    pd.DataFrame
+        Query results in pandas DataFrame
+
+    See Also
+    --------
+    `func`:search.query:
```
mlshapiro, line 99:
```suggestion
    :func:`search.query`
```

pycontrails/datalib/landsat.py
```diff
@@ -0,0 +1,562 @@
...
+        By default, returns imagery base URL and sensing time.
+
+    Returns
+    -------
+    pd.DataFrame
+        Query results in pandas DataFrame
+
+    See Also
+    --------
+    :func"`search.intersect`
```
mlshapiro, line 136:
```suggestion
    :func:`search.intersect`
```

mlshapiro (approved):
Really impressive work here! 

I made a few organizational changes and rebased off main. Primarily to avoid having a submodule level `datalib.leo`. 

We can discuss when you're back if you disagree with these changes.

Notebooks work as expected, although the initial load on `handler.get()` calls take a while. May be worth throwing a comment or quote in there to share that we expect these calls to take a while the first time.

thabbott:
I think your re-organization makes sense! Moving the generic `search` and `query` functions to `_leo_utils` should help make it clear to users that they (in almost all cases) want to use e.g. `landsat.search` and not `_leo_utils.search`.

I added some quotes about `get` calls taking a while to notebooks in https://github.com/contrailcirrus/pycontrails/pull/200/commits/8dcd483bf3c3254c28aba6a21cc3b44d0ce6d22b. (They've been pretty quick for me, but that's been running on a VM with fast transfer speeds from google cloud buckets. I suspect you were running locally?)

pycontrails/datalib/_leo_utils/search.py
thabbott: To be honest I didn't realize these methods exist. Yes, we should reconcile them with what I implemented!

In https://github.com/contrailcirrus/pycontrails/pull/200/commits/ce0d49475a123b1006898c7f6f38332c6dc66cc5 I
- modified `Flight.to_geojson_multilinestring` to allow multiple antimeridian crossings and to make the grouping `key` parameter optional, and
- updated `track_to_geojson` to use `Flight.to_geojson_multilinestring` for antimeridian splitting. I still have to do some postprocessing work in `track_to_geojson`, since I don't want altitude coordinates and I want to pad linestrings so they extend all the way to the antimeridian before and after crossings, but this at least means we don't have two independent versions of antimeridian splitting logic to maintain.

This introduces a breaking change, but only in the return type of `flight._antimeridian_index` (was `int`, now is `list[int]`). Let me know what you think!

(Where if anywhere is `Flight.to_geojson_multilinestring` used? As far I can tell it's not used in pycontrails except in tests, and now in `track_to_geojson`.)

PR #202, opened 2024-06-07
mypy updates

mlshapiro:
Closes #XX

#### Internals

- Update dependency version of mypy
- Remove outdated `type: ignore` statements

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer


mlshapiro:
I think we should pin mypy higher, but the changes to the files don't work...I'm not going to think about this any more

PR #204, opened 2024-06-14
Exclude netcdf4 1.7.0 in dependencies

thabbott:
Unit tests have been failing non-deterministically since the release of [netCDF4 version 1.7.0](https://pypi.org/project/netCDF4/1.7.0/).

To reproduce failures, run
```bash
while pytest tests/unit/test_ecmwf.py; do; done
```
with netCDF4 1.7.0 installed.

For failures in CI, see e.g. https://github.com/contrailcirrus/pycontrails/actions/runs/9519726634.

Downgrading to netCDF4 1.6.5 appears to resolve this problem.

#### Fixes

- Exclude netCDF version 1.7.0, which causes non-deterministic failures in unit tests primarily due to segmentation faults.

## Tests

- [x] CI tests pass

## Reviewer

> @mlshapiro

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ ecmwf = [
     "cfgrib>=0.9",
     "eccodes>=1.4",
     "ecmwf-api-client>=1.6",
-    "netcdf4>=1.6.1,!=1.7.0",
+    "netcdf4>=1.6.1,<1.7.0",
     "platformdirs>=3.0",
     "requests>=2.25",
     "lxml>=5.1.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ ecmwf = [
     "cfgrib>=0.9",
     "eccodes>=1.4",
     "ecmwf-api-client>=1.6",
-    "netcdf4>=1.6.1",
+    "netcdf4>=1.6.1,!=1.7.0",
     "platformdirs>=3.0",
     "requests>=2.25",
     "lxml>=5.1.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -98,7 +98,7 @@ ecmwf = [
     "cfgrib>=0.9",
     "eccodes>=1.4",
     "ecmwf-api-client>=1.6",
-    "netcdf4>=1.6.1",
+    "netcdf4>=1.6.1,!=1.7.0",
```
mlshapiro, line 101: I believe the syntax you want it 

```suggestion
    "netcdf4>=1.6.1,<1.7.0",
```


unless you want it to install 1.7.1 or higher when it comes out?

mlshapiro (approved):

pyproject.toml
thabbott: I used !=1.7.0 knowing that it would mean that 1.7.1 gets installed.

I don't have a strong opinion about whether it's better to allow 1.7.1 to get installed and potentially deal with the same problem again, or better to pin to <1.7.0 and have to remember to relax the constraint at some point in the future when this issue gets fixed.

I'll go with the syntax you suggested unless you chime in again in the next hour or so!

mlshapiro: I would just stick to 1.6.x (so with my suggested edit) for now

zebengberg: I've been using 1.7.0 locally, and I've also been getting segfaults.

I think netcdf made their release in large part to support numpy 2.0, which itself is scheduled for release this Sunday. Because of this deadline, I believe netcdf may make another release shortly.

thabbott: Hmm, ok. I want to merge this because the segfaults are stopping me from finishing up some other PRs, but if netcdf makes another release that fixes the issue I can change the constraint to install it.

PR #212, opened 2024-06-28
Bugfix/persistent contrails formed

trdean1:
#### Fixes

Fixes bug in `output_format.py` so that `total_persistent_contrails_formed` is computed in a manner that is consistent for the convention of computing EF.

Zero EF = no contrail formed (or if it did it was less than one segment length)
NaN EF = missing input to compute contrail output

Both cases should be excluded from `total_persistent_contrails_formed`.

## Tests

- [ ] CI tests pass

### DIFF
--- a/pycontrails/models/cocip/output_formats.py
+++ b/pycontrails/models/cocip/output_formats.py
@@ -217,7 +217,7 @@ def contrail_flight_summary_statistics(flight_waypoints: GeoVectorDataset) -> pd
     )
 
     flight_waypoints["persistent_contrail_length"] = np.where(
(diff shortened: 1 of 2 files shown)


thabbott:
Maybe worth adding a test that the number of waypoints with an RF value is always larger than the number of waypoints that form persistent contrails?
```python
(flight_summary["total_persistent_contrails_formed"] == 0.0).sum() <= (flight_summary["mean_lifetime_rf_net"].isna()).sum()
```
This way our tests still assert something about the `mean_lifetime_rf_net` summary statistic.

Otherwise looks good to me!

PR #213, opened 2024-07-01
Enable file locks in `MetDataSource.open_dataset`

thabbott:
## Changes

These changes prevent segmentation faults when reading and writing multi-file datasets with recent netCDF4 versions (see https://github.com/contrailcirrus/pycontrails/pull/204) and allow us to remove upper limits on netCDF4 and numpy versions.

These changes revert a change (https://github.com/contrailcirrus/pycontrails/pull/68) made to address the issue described in https://github.com/pydata/xarray/issues/4406. That issue remains open, so this change may resurface problems with dask threadlocks.

### Breaking changes

- Remove `lock=False` as a default keyword argument to `xr.open_mfdataset` in the `MetDataSource.open_dataset` method. This reverts a change from [v0.44.1](https://github.com/contrailcirrus/pycontrails/releases/tag/v0.44.1) and prevents segmentation faults when using recent versions of [netCDF4](https://pypi.org/project/netCDF4/) (1.7.0 and above).

### Internals

- Remove upper limits on netCDF4 and numpy versions.

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+### Breaking changes
+
+- Remove `lock=False` as a default keyword argument to `xr.open_mfdataset` in the `MetDataSource.open_dataset` method. This reverts a change from [v0.44.1](https://github.com/contrailcirrus/pycontrails/releases/tag/v0.44.1) and prevents segmentation faults when using recent versions of [netCDF4](https://pypi.org/project/netCDF4/) (1.7.0 and above).
+- GOES imagery is now loaded from a temporary file on disk rather than directly from memory when using `GOES.get` without a cachestore.
+
+### Internals
+
+- Remove upper limits on netCDF4 and numpy versions.
+- Remove h5netcdf dependency.
+- Update doctests with numpy 2 scalar representation (see [NEP 51](https://numpy.org/neps/nep-0051-scalar-representation.html)).
```
zebengberg, line 14: So will the doctests now fail with numpy<2?

pycontrails/datalib/goes.py
```diff
@@ -564,6 +564,14 @@ def _get_without_cache(self, time: datetime.datetime) -> xr.DataArray:
         return da
 
 
+def _load_via_tempfile(data: bytes) -> xr.Dataset:
+    """Load xarray dataset via temporary file."""
+    with tempfile.NamedTemporaryFile(buffering=0) as tmp:
+        tmp.write(data)
+        ds = xr.load_dataset(tmp.name)
+    return ds
```
zebengberg, line 572: You can save a single line of code with:

```python
    with tempfile.NamedTemporaryFile(buffering=0) as tmp:
        tmp.write(data)
        return xr.load_dataset(tmp.name)
```

(the context manager will still exit even if there's a return in there)

zebengberg (approved):
Thanks Tristan!

CHANGELOG.md
thabbott: Yes. I added a check in the Makefile (https://github.com/contrailcirrus/pycontrails/pull/213/commits/d39d70e98fea4355fcb7d11191581c33f4d8cbab) to make this requirement explicit. 

Trying to keep doctests compatible with numpy 1 and 2 seems tricky.... You can get the same output for scalars by using `print`, but that doesn't work for scalars wrapped in some other object:
```
>>> np.float64(1.0)
1.0              # numpy 1
np.float64(1.0)  # numpy 2
>>> print(np.float64(1.0)
1.0  # numpy 1
1.0  # numpy 2
>>> print((np.float64(1.0), np.float64(2.0))
(1.0, 2.0)                          # numpy 1
(np.float64(1.0), np.float64(2.0))  # numpy 2
```

Happy to discuss this more if you don't think this seems like the right solution!

pycontrails/datalib/goes.py
thabbott: https://github.com/contrailcirrus/pycontrails/pull/213/commits/44911214d71a9af2b39043f84ec6076cdccbf6ac

PR #214, opened 2024-07-03
localize pycontrail GOES import

nickmasson:
## Changes
Small refactor to localize the import of the pycontrails datalib GOES module.

In its current manifestation, the GOES module is imported whenever `output_formats.py` is used (which is prevalent).
The GOES module requires [optional dependencies](https://github.com/contrailcirrus/pycontrails/blob/98008ed98690f62d72027596cf23a5930c9a7c61/pyproject.toml#L124), however, that most clients would prefer not to import by default.


zebengberg (approved):
Thank you!

PR #216, opened 2024-07-08
Fix/float promotion

zebengberg:
- Document pycontrails installation from conda-forge.
- Ensure the `MetDataset` vertical coordinates `"air_pressure"` and `"altitude"` have the correct dtype.

## Tests

- [x] CI tests pass


pycontrails/core/met.py
```diff
@@ -2625,25 +2627,47 @@ def _add_vertical_coords(data: XArrayType) -> XArrayType:
...
+        pass
     else:
-        dtype = data.dtype
+        if air_pressure.dtype == dtype and altitude.dtype == dtype:
+            return data
+        # At least one of 'air_pressure' or 'altitude' has the wrong dtype
+        data.coords["air_pressure"] = air_pressure.astype(dtype, copy=False)
+        data.coords["altitude"] = altitude.astype(dtype, copy=False)
+        return data
+
```
thabbott, line 2649: Is this try-else block really necessary?

I think you could simplify the code by deleting it and changing the if-else blocks below it to
```python
if "air_pressure" not in coords:
   ...
elif data.coords["air_pressure"].dtype != dtype:
   ...
```
Maybe there is a corner case I haven't thought of, though.

tests/unit/test_sac_issr.py
```diff
@@ -65,6 +66,7 @@ def test_ISSR_met_source(met_issr: MetDataset) -> None:
     # undergo humidity scaling. This is why the counts below are lower.
     assert "humidity_scaling" not in out2.attrs
     assert isinstance(out2, MetDataset)
+    all(v == np.float32 for v in out2.data.dtypes.values())
```
thabbott, line 69:
```suggestion
    assert all(v == np.float32 for v in out2.data.dtypes.values())
```

thabbott (approved):
Looks good! Left a few minor comments; the main one is related to cleaning up _add_vertical_coords.

pycontrails/core/met.py
mlshapiro: I agree - there may be a slight performance improvement with try-except, but for clarity, I think we should handle key addition and dtype all in the same place

thabbott: https://github.com/contrailcirrus/pycontrails/pull/216/commits/e48adb2a37b3cfa06d551f8546a9c39493f342c9

PR #220, opened 2024-07-18
Feature/isa temperature below met

zebengberg:
Add experimental `interpolation_fill_low_alt_with_isa_temperature` model parameter. When set to `True`, models with `GeoVectorDataset` sources will fill points below the lowest altitude in the ``met["air_temperature]`` data with the ISA temperature. This is useful for aircraft performance calculations when the met data does not extend to the surface. In this case, we can still estimate fuel flow and other performance metrics through the entire climb phase. By default, this parameter is set to `False`.

## Tests

- [x] CI tests pass

@thabbott 

The main motivation for this PR is to compute aircraft performance estimates (most notably, fuel flow and aircraft mass) below the lowest-altitude met level. To do this, we need air temperature (which we approximate with ISA temperature). But we also need u and v wind to compute the true airspeed. How should we handle these unknown winds? One crude option is to assume no wind below the lowest-altitude met level (in other words, set true airspeed to true groundspeed).


tests/unit/test_leo.py
```diff
@@ -15,6 +16,7 @@
 from tests import BIGQUERY_ACCESS, OFFLINE
 
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
+PYCONTRAILS_SKIP_LEO_TESTS = bool(os.getenv("PYCONTRAILS_SKIP_LEO_TESTS"))
```
zebengberg, line 19: @thabbott I added this because the unit tests were taking too long for me to run locally. I'm happy to revert if you think this is a bad idea. (I can always set a `GITHUB_ACTIONS` environment variable as a hack.)

mlshapiro: This makes me wonder at what point we start using a conf file....

mlshapiro:
I'll think about this more today. 

We need some kind of solution for this. I think the proposed ISA approach will work for temperature since the AP is less sensitive to temperature. But the sensitivity to wind may be too high to assume true_airspeed is true_groundspeed (Can we check this?)

Its clear we need the gridded data for any kind of parcel tracking in the contrail models, but for the aircraft performance calculations, we only need meteorological data pointwise. Would it make sense to integrate with some kind of point-wise meteo api for the aircraft performance? I'm thinking of [OpenMeteo](https://open-meteo.com) or similar. We'd pass in points and request the data at those points for the forecast perscribed? Unfortunately a bit of a hack.

tests/unit/test_leo.py
thabbott: I think it makes sense to have something besides GITHUB_ACTIONS that you can use to skip tests locally. Maybe Marc has ideas for a different way to implement conditional skipping? We could bring this up at standup to make sure there's not a better approach.

tests/unit/test_leo.py
```diff
@@ -249,11 +251,13 @@ def landsat_base_url() -> str:
...
 @pytest.fixture()
-def landsat_cachestore() -> cache.DiskCacheStore:
+def landsat_cachestore() -> Generator[cache.DiskCacheStore, None, None]:
     """Clearable cache for Landsat data."""
     cache_root = cache._get_user_cache_dir()
     cache_dir = f"{cache_root}/landsat-unit-test"
-    return cache.DiskCacheStore(cache_dir=cache_dir, allow_clear=True)
+    cachestore = cache.DiskCacheStore(cache_dir=cache_dir, allow_clear=True)
+    yield cachestore
+    cachestore.clear()
```
thabbott, line 260: This is a nice change!

There is probably some room to reduce the time it takes to run the LEO test suite by letting data persist in the cache across tests. For example, test_landsat_get_reflective_bands currently downloads each band three times (once for each "processing" value). I structured the tests this way so they'd be as independent as possible, but if they're painfully slow for you we could consider changes that reduce independence but improve performance.

pycontrails/core/aircraft_performance.py
```diff
@@ -543,3 +586,54 @@ class AircraftPerformanceGridData(Generic[ArrayOrFloat]):
...
+    met_level_max : float
+        The maximum level in the met data, [:math:`hPa`].
+    """
+    air_temperature = vector["air_temperature"]
+    is_nan = np.isnan(air_temperature)
+    low_alt = vector.level > met_level_max
+    cond = is_nan & low_alt
+
+    t_isa = vector.T_isa()
+    air_temperature[cond] = t_isa[cond]
```
thabbott, line 614: You could come up with slightly more reasonable temperature profiles by using the ISA lapse rate but matching temperatures at the lowest level of the met data. That would make the implementation significantly more complicated (you'd need some information about met data temperatures in this function), and I'm not sure how much it would actually improve fuel burn estimates.... So I don't have a strong opinion about whether this is worth your time.

thabbott (approved):
This looks good to me! Filling u and v with zeros seems like the right decision. (I think the primary goal is to avoid having zero fuel burn below the bottom of the met data, and the error from setting u and v to zero should be small by comparison.) I left one comment about a possible improvement to temperature filling--up to you whether that seems worthwhile.

Edit: Marc seems less confident than I am that setting u = v = 0 is a reasonable assumption... so maybe it is worth validating a bit more carefully.

tests/unit/test_leo.py
zebengberg: Yeah ... it's really just the download that is slow. We could take a similar approach that we use for some the notebooks, and not clear the cache between test runs.

I'll leave as is for now.

pycontrails/core/aircraft_performance.py
```diff
@@ -39,6 +40,17 @@ class AircraftPerformanceParams(ModelParams):
...
+    #: level with ISA temperature when interpolating "air_temperature" or "t".
+    #: If the ``met`` data is not provided, the entire air temperature array
+    #: is approximated with the ISA temperature.
+    fill_low_altitude_with_isa_temperature: bool = False
+
+    #: Experimental. If True, fill waypoints below the lowest altitude met
+    #: level with zero wind when computing true airspeed. In other words,
+    #: approximate low-altitude true airspeed with the ground speed.
+    fill_low_altitude_with_zero_wind: bool = False
+
```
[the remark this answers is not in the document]
zebengberg: 1958cefade7c1d49eb688f9f6b465865c7937940

pycontrails/core/aircraft_performance.py
zebengberg: I thought about doing something like this. Without having a good source of ground truth (for example, ERA5 data all the way down to the surface), I decided not to pursue anything more sophisticated.

zebengberg:
> This looks good to me! Filling u and v with zeros seems like the right decision. (I think the primary goal is to avoid having zero fuel burn below the bottom of the met data, and the error from setting u and v to zero should be small by comparison.) I left one comment about a possible improvement to temperature filling--up to you whether that seems worthwhile.
> 
> Edit: Marc seems less confident than I am that setting u = v = 0 is a reasonable solution... so maybe it is worth validating a bit more carefully.

I poked at this a bit more (nothing particularly rigorous). Below is a scatter plot of ~140000 waypoints in the climb phase from a fleet of ~6000 flights. I ran them all through `PSFlight` with these two new flags enabled. The lowest-altitude met level in my weather data is around 22000ft. I used the red vertical line to show this in my plot. The fuel flow to the left of this line is computing using TAS = groundspeed and temperature = ISA. The fuel flow to the right is computing by interpolating against the met data.

The predictions don't go haywire to the left of the line, but you can see that the spread in the fuel flow distribution is larger compared with the met-interpolated points.

Screenshot 2024-07-24 at 3 09 47 PM

zebengberg:
@thabbott I'm going to merge this branch, but we can hold off on making a release until we're a little more certain that this gives an improvement to our aircraft performance estimates.

PR #221, opened 2024-07-24
A few performance improvements when running CoCiP with a Fleet

zebengberg:
### Fixes

- Fix the `step_threshold` in `Flight._altitude_interpolation`. This correction is only relevant when passing in a non-default value for `freq` in `Flight.resample_and_fill`.
- Fix the `VectorDataset.__eq__` method to check for the same keys. Previously, if the other dataset had a superset of the instance keys, the method may still return `True`.

### Internals

- Improve the runtime performance of `Fleet.to_flight_list`. For a large `Fleet`, this method is now 5x faster. The plot below shows my benchmarking.
Screenshot 2024-07-22 at 6 22 57 PM

- Improve the runtime performance and memory footprint of `Cocip._bundle_results`. When running `Cocip` with a large `Fleet` source, `Cocip.eval` is now slightly faster and uses much less memory.

## Tests

- [ ] CI tests pass

### DIFF
--- a/pycontrails/core/fleet.py
+++ b/pycontrails/core/fleet.py
@@ -226,28 +226,29 @@ class Fleet(Flight):
         return len(self.fl_attrs)
 
     def to_flight_list(self, copy: bool = True) -> list[Flight]:
-        """De-concatenate merged waypoints into a list of Flight instances.
+        """De-concatenate merged waypoints into a list of :class:`Flight` instances.
 
         Any global :attr:`attrs` are lost.
 
         Parameters
         ----------
         copy : bool, optional
-            If True, make copy of each flight instance in `seq`.
+            If True, make copy of each :class:`Flight` instance.
 
         Returns
         -------
         list[Flight]
-            List of Flights in the same order as was passed into the `Fleet` instance.
+            List of Flights in the same order as was passed into the ``Fleet`` instance.
         """
-
-        # Avoid self.dataframe to purposely drop global attrs
-        tmp = pd.DataFrame(self.data, copy=copy)
-        grouped = tmp.groupby("flight_id", sort=False)
-
+        indices = self.dataframe.groupby("flight_id", sort=False).indices
         return [
-            Flight(df, attrs=self.fl_attrs[flight_id], fuel=self.fuel, copy=copy)
-            for flight_id, df in grouped
+            Flight(
+                data=VectorDataDict({k: v[idx] for k, v in self.data.items()}),
+                attrs=self.fl_attrs[flight_id],
+                copy=copy,
+                fuel=self.fuel,
+            )
+            for flight_id, idx in indices.items()
         ]
 
     ###################################
--- a/pycontrails/core/flight.py
+++ b/pycontrails/core/flight.py
@@ -2234,16 +2234,14 @@ def segment_rocd(
     if air_temperature is None:
         return out
 
-    else:
-        altitude_m = units.ft_to_m(altitude_ft)
-        T_isa = units.m_to_T_isa(altitude_m)
+    altitude_m = units.ft_to_m(altitude_ft)
+    T_isa = units.m_to_T_isa(altitude_m)
 
-        T_correction = np.empty_like(altitude_ft)
-        T_correction[:-1] = (0.5 * (air_temperature[:-1] + air_temperature[1:])) / (
-            0.5 * (T_isa[:-1] + T_isa[1:])
-        )
-        T_correction[-1] = np.nan
-        return T_correction * out
+    T_correction = np.empty_like(altitude_ft)
+    T_correction[:-1] = (air_temperature[:-1] + air_temperature[1:]) / (T_isa[:-1] + T_isa[1:])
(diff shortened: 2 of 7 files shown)


pycontrails/core/flight.py
```diff
@@ -2234,16 +2234,14 @@ def segment_rocd(
...
+    T_isa = units.m_to_T_isa(altitude_m)
 
-        T_correction = np.empty_like(altitude_ft)
-        T_correction[:-1] = (0.5 * (air_temperature[:-1] + air_temperature[1:])) / (
-            0.5 * (T_isa[:-1] + T_isa[1:])
-        )
-        T_correction[-1] = np.nan
-        return T_correction * out
+    T_correction = np.empty_like(altitude_ft)
+    T_correction[:-1] = (air_temperature[:-1] + air_temperature[1:]) / (T_isa[:-1] + T_isa[1:])
```
zebengberg, line 2241: Besides un-indenting one level, I also removed the 0.5 from the numerator and denominator here -- it looked more readable to me.

zebengberg: We could also use `np.diff` as well.

pycontrails/models/cocip/cocip.py
```diff
@@ -1166,49 +1166,43 @@ def _bundle_results(self) -> None:
         # ---
         # Create contrail dataframe (self.contrail)
         # ---
-        dfs = [contrail.dataframe for contrail in self.contrail_list]
-        dfs = [df.assign(timestep=t_idx) for t_idx, df in enumerate(dfs)]
-        self.contrail = pd.concat(dfs)
+        for t_idx, contrail in enumerate(self.contrail_list):
+            contrail["timestep"] = np.full(contrail.size, t_idx)
+
+        self.contrail = GeoVectorDataset.sum(self.contrail_list).dataframe
```
zebengberg, line 1172: This change essentially replaces `pd.concat` with `np.concatenate`. I've found the later to be a bit faster but also use a lot less memory.

thabbott (approved):
This looks great to me!

One question: does our definition of a "breaking change" just include changes to intended behavior, or also bug fixes that change behavior for cases affected by the bug? (The `step_threshold` changes are in the second category, and I'm not sure if we should note them in the breaking changes section of the changelog.)

pycontrails/core/flight.py
zebengberg: The current implementation would be a hair faster. [`np.diff` does exactly](https://github.com/numpy/numpy/blob/893215af7f97105fa2dccb5bb560bc322dda6c77/numpy/lib/_function_base_impl.py#L1508-L1515) what we do here with a little more setup and tear down.

pycontrails/core/fleet.py
```diff
@@ -226,28 +226,29 @@ def n_flights(self) -> int:
...
-        # Avoid self.dataframe to purposely drop global attrs
-        tmp = pd.DataFrame(self.data, copy=copy)
-        grouped = tmp.groupby("flight_id", sort=False)
-
+        indices = self.dataframe.groupby("flight_id", sort=False).indices
         return [
-            Flight(df, attrs=self.fl_attrs[flight_id], fuel=self.fuel, copy=copy)
-            for flight_id, df in grouped
+            Flight(
+                data=VectorDataDict({k: v[idx] for k, v in self.data.items()}),
```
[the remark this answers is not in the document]
zebengberg: No great reason -- using a `VectorDataDict` allow us to [avoid one more redundant dictionary copy](https://github.com/contrailcirrus/pycontrails/blob/ac27c89eb73191c9d529b4662d96d854efa79356/pycontrails/core/vector.py#L304-L312) (which is cheap here compared to the groupby, so I'm splitting hairs).

pycontrails/core/flight.py
```diff
@@ -1506,25 +1506,25 @@ def length_met(self, key: str, threshold: float = 1.0) -> float:
...
+        >>> fl["air_temperature"] = fl.intersect_met(met["air_temperature"])
         >>> fl["air_temperature"]
-        array([235.94657007, 235.95766965, 235.96873412, ..., 234.59917962,
+        array([235.94657007, 235.55745645, 235.56709768, ..., 234.59917962,
                234.60387402, 234.60845312])
 
         >>> # Length (in meters) of waypoints whose temperature exceeds 236K
         >>> fl.length_met("air_temperature", threshold=236)
-        np.float64(4132178.159...)
+        np.float64(3589705.998...)
```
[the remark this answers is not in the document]
zebengberg: Yes, exactly -- it's the change to the `step_threshold`.

zebengberg:
> One question: does our definition of a "breaking change" just include changes to intended behavior, or also bug fixes that change behavior for cases affected by the bug? (The `step_threshold` changes are in the second category, and I'm not sure if we should note them in the breaking changes section of the changelog.)

You're absolutely right -- we often blur this line. In this case, I think it's enough to note the change just in the fix category ... but it's definitely a matter of opinion.

PR #227, opened 2024-08-01
defer matplotlib import

### LINKED ISSUE #226: minor (pseudo-regression) in newer pycontrails releases
**Description**

Newer releases of pycontrails include an import of matplotlib in the `output_formats` of `models.cocip`.
See [here](https://github.com/contrailcirrus/pycontrails/blob/dd9b453918c490819af3d2adcd06c898f1be827e/pycontrails/models/cocip/output_formats.py#L27).

Matplot lib is only referenced in the [`compare_cocip_with_goes` method](https://github.com/contrailcirrus/pycontrails/blob/dd9b453918c490819af3d2adcd06c898f1be827e/pycontrails/models/cocip/output_formats.py#L2098).

Matplotlib is an optional dependency under [`vis`](https://github.com/contrailcirrus/pycontrails/blob/dd9b453918c490819af3d2adcd06c898f1be827e/pyproject.toml#L144)

**Change Request**
The `import matplotlib.pyplot as plt` should be localized (moved) inside the code block for this method definition.
This will avoid forcing the user to import matplotlib when using any of the other methods in `output_formats` (i.e. for users who are not making use of the optional `vis` dependencies and methods).

**Details**

- Version: [e.g. 0.52.2]
- Module: [e.g. models.cocip.output_formats]
- OS: linux
- Reporter: N. Masson

zebengberg:
Closes #226 

## Tests

- [ ] CI tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.52.3 (unreleased)
+
+### Fixes
+
+- Defer import of `matplotlib` in `models.cocip.output_formats`.
+
 ## v0.52.2
 
 ### Breaking changes
--- a/pycontrails/models/cocip/output_formats.py
+++ b/pycontrails/models/cocip/output_formats.py
@@ -24,7 +24,6 @@ import pathlib
 import warnings
 from collections.abc import Hashable
 
-import matplotlib.pyplot as plt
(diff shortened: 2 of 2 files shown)


nickmasson (approved):
Thanks @zebengberg!

Issue #232, opened 2024-08-14
ERA5 downloading and preprocessing with new CDSAPI-BETA


Reyrem:
**ERA5 downloading and preprocessing with new CDSAPI-BETA**

CDSAPI changed plateform, the ancien one will be decommissioned beginning of september. The old plateform is now subject to very long waiting time. 

The time coordinate in the files returned by the new api is now called "valid_time". The level dimension is now called "pressure_level". This causes disruption when using pycontrails to download, preprocess and cache weather data. 

It seems that the latitude and longitude dimensions might be affected as well. 

**Steps to Reproduce**

1. follow ecwmf procedure to migrate to cdsapi beta
2. run pycontrail era5 downloading methods as usual 

**Additional Notes**

A quick and dirty fix could be to go  to <path_to_environments>/envs/pycontrail/lib/python3.10/site-packages/pycontrails/datalib/ecmwf/era5.py  and modify the _download_file function by adding these two lines : 

self.cds.retrieve(self.dataset, request, cds_temp_filename)

\# open file, edit, and save for each hourly time step
ds = stack.enter_context(
    xr.open_dataset(cds_temp_filename, engine=metsource.NETCDF_ENGINE)
)
**ds = ds.swap_dims ({'valid_time' : 'time', 'pressure_level' : 'level'})
ds = ds.rename_vars({'valid_time' : 'time', 'pressure_level' : 'level'})**

\# run preprocessing before cache
ds = self._preprocess_era5_dataset(ds)

and for some reason I also needed to change a line in <path_to_environments>/envs/pycontrail/lib/python3.10/site-packages/pycontrails/datalib/ecmwf/common.py, in function cache_dataset

line 98 goes from             
cache_path = self.create_cachepath(pd.Timestamp(t).to_pydatetime())
to
cache_path = self.create_cachepath(pd.Timestamp(ds_t.time.values[0]).to_pydatetime())

Reyrem:
if you are trying to download files from the surface weather dataset, maybe you will have a problem because there might not be a dimension "level".
 
try adding  :
if "level" not in ds.dims and len(self.pressure_levels) == 1:
            ds = ds.expand_dims(level=self.pressure_levels)

or something similar just after opening the ds xarray  in era5.py

Hope it helps. otherwise, I saw people opening another ticket related to this topic :
#233

Brian75321:
[...]

PR #233, opened 2024-08-15
Ensure CDS-beta compatibility, improve aCCF calculation performance, and fix aCCF parameters

jsmretschnig:
Soon ERA5 will only be accessible via the new API, however some variable names are different. This PR renames those variables.

The CLIMaCCF interface is not implemented correctly: one parameter was wrong and the assumption that all aCCFs are returned with unit K/kg(fuel) was also not correct.

Furthermore, the result from CLIMaCCF does not need to be sorted, as it is already sorted.

### DIFF
--- a/pycontrails/core/met.py
+++ b/pycontrails/core/met.py
@@ -229,7 +229,7 @@ class MetBase(ABC, Generic[XArrayType]):
         self._validate_latitude()
         self._validate_transpose()
 
-    def _preprocess_dims(self, wrap_longitude: bool) -> None:
+    def _preprocess_dims(self, wrap_longitude: bool, sort: bool = True) -> None:
         """Confirm DataArray or Dataset include required dimension in a consistent format.
 
         Expects DataArray or Dataset to contain dimensions ``latitude`, ``longitude``, ``time``,
@@ -250,6 +250,8 @@ class MetBase(ABC, Generic[XArrayType]):
         ----------
         wrap_longitude : bool
             If True, ensure longitude values cover the interval ``[-180, 180]``.
+        sort : bool
+            If True, coordinates are sorted with ascending order.
 
         Raises
         ------
@@ -266,7 +268,8 @@ class MetBase(ABC, Generic[XArrayType]):
                 self.data[coord] = arr.astype(COORD_DTYPE)
 
         # Ensure time is np.datetime64[ns]
-        self.data["time"] = self.data["time"].astype("datetime64[ns]", copy=False)
+        if sort:
+            self.data["time"] = self.data["time"].astype("datetime64[ns]", copy=False)
 
         # sortby to ensure each coordinate has ascending order
         self.data = self.data.sortby(list(self.dim_order), ascending=True)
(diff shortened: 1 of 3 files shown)


jsmretschnig:
Also connected to https://github.com/contrailcirrus/pycontrails/issues/232 as I just found out :)

pycontrails/core/met.py
```diff
@@ -266,7 +268,8 @@ def _preprocess_dims(self, wrap_longitude: bool) -> None:
                 self.data[coord] = arr.astype(COORD_DTYPE)
 
         # Ensure time is np.datetime64[ns]
-        self.data["time"] = self.data["time"].astype("datetime64[ns]", copy=False)
+        if sort:
+            self.data["time"] = self.data["time"].astype("datetime64[ns]", copy=False)
```
zebengberg, line 272: I don't understand this ... here, we're ensuring the time coordinate has the right dtype, so how is this related to `sort`?

pycontrails/core/met.py
```diff
@@ -280,7 +283,7 @@ def _preprocess_dims(self, wrap_longitude: bool) -> None:
             # Only shift if necessary
             if np.any(lon >= 180.0) or np.any(lon < -180.0):
                 self.data = shift_longitude(self.data)
-            else:
+            elif sort:
```
zebengberg, line 286: We already call `self.data.sortby(...)` 11 lines above this. Why make this sort conditional? There is also a `copy` parameter in the `MetDataset` constructor. If that is set to False, then I think we avoid unnecessary sorting.

pycontrails/models/accf.py
```diff
@@ -267,7 +267,7 @@ def eval(
         aCCFs, _ = clim_imp.get_xarray()
 
         # assign ACCF outputs to source
-        maCCFs = MetDataset(aCCFs)
+        maCCFs = MetDataset(aCCFs, sort=False)
```
zebengberg, line 270: I wonder if you could achieve what you're doing here by manually making any tweaks to `aCCFs` first, then changing this line to `maCCFs = MetDataset(aCCFs, copy=False)`.

pycontrails/datalib/ecmwf/era5.py
```diff
@@ -488,6 +488,11 @@ def _download_file(self, times: list[datetime]) -> None:
 
             # run preprocessing before cache
             ds = self._preprocess_era5_dataset(ds)
+            # renaming to comply with new CDS-beta
+            # date of shutdown of the previous API: September 3rd, 2024
+            ds = ds.rename({"valid_time": "time"}).drop_vars("number")
+            if len(self.pressure_levels) > 1:
+                ds = ds.rename({"pressure_level": "level"})
```
zebengberg, line 495: Thank you! Do you know if this is backwards compatible with the legacy CDS client?

zebengberg (commented):
Thank you so much @jsmretschnig (and @Reyrem)!

I left a few comments for you.

pycontrails/core/met.py
jsmretschnig: You're right, I mixed up the line here. The if-statement is supposed to be two lines below. But yeah let me check if `copy=False` achieves the same, then I can also remove the sort parameter again.

zebengberg:
@jsmretschnig I updated the `ERA5` interface in #234, so the latest pycontrails release can now handle ERA5 data from the new CDS-Beta as well as the legacy CDS server. Feel free to open another PR with your CLIMaCCF fixes!

PR #236, opened 2024-08-27
Feature/ruff format

zebengberg:
## Internal changes

- Use `ruff` to format the codebase in place of `black`.
- Run `ruff` linting and formatting over the notebook examples in the documentation.
- Update development documentation with new links.

## Tests

- [x] CI tests pass

## Reviewer

> Select @ reviewer


Makefile
```diff
@@ -71,14 +71,20 @@ dev-pycontrails-bada:
...
+	ruff format pycontrails tests docs
+
+ruff:
+	@echo -e "$(COLOR_YELLOW)The 'ruff' target is deprecated. Use 'make lint' instead.\n"
+	make lint
 
-black-check:
-	black pycontrails tests --check
+black:
+	@echo -e "$(COLOR_YELLOW)The 'black' target is deprecated. Use 'make format' instead.\n"
```
mlshapiro, line 86: Is there any reason to keep these? I think we are the only ones with workflows around them

pyproject.toml
```diff
@@ -203,13 +202,6 @@ show_error_codes = true
 # strict_equality = true
 # # --strict end
 
-# black code formatting
-# https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file
-[tool.black]
-line-length = 100
```
mlshapiro, line 209: is the default line-length for ruff 100? I don't see this configured below

pyproject.toml
```diff
@@ -203,13 +202,6 @@ show_error_codes = true
...
-preview = true
-required-version = "24.8.0"
```
mlshapiro, line 211: is there anything similar for ruff? This is kind of helpful for footgun stuff

mlshapiro (approved):

pyproject.toml
zebengberg: The default is 88, but we'd previously specified this as 100 under `tool.ruff` which is why you don't see a diff there.

pyproject.toml
zebengberg: Unfortunately no. We've pinned the ruff version in the dev dependencies, but the ruff configuration doesn't confirm it. As ruff matures more, I think breaking changes will be less common.

PR #245, opened 2024-09-27
Fix/cocip grid dt

zebengberg:
Closes #243 

### Fixes

- Fix the integration time step in `CocipGrid.calc_evolve_one_step`. The previous implementation assumed a time interval of `params["dt_integration"]`. This may not be the case for all `source` parameters (for example, this could occur if running `CocipGrid` over a collection of ADS-B waypoints).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 # Changelog
 
-## (unreleased)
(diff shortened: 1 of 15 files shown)


thabbott (approved):
Looks good to me!

PR #252, opened 2024-10-07
Remove deprecations, remove crs, rename ARCOERA5

### LINKED ISSUE #151: Deprecate or remove GeoVectorDataset.attrs["crs"]
Generally, pycontrails assumes all geospatial data is provided in WGS84. In the `GeoVectorDataset` docstring, we write:

> Expect latitude-longitude CRS in WGS 84.

The `GeoVectorDataset` interface gives an option to set a custom CRS through its `attrs`. This attribute is not consistently used in geospatial calculations. As a concrete example, `Flight.length` and `Flight.segment_length` both check the crs attribute, whereas `Flight.segment_angle` and `Flight.segment_groundspeed` do not. In the cases where the CRS is checked, an error is raised when the data is not WGS84 (the default).

The `GeoVectorDataset.attrs["crs"]` can likely be removed without impacting the end user. It would result in a slightly cleaner code base, slight performance improvements, and simplify some patterns (for example, `Fleet` constructions).

### LINKED ISSUE #180: Rename ARCOERA5 to ERA5ARCO
**Description**

Rename the `ARCOERA5` class to `ERA5ARCO` for consistency with datalibs for other ERA5 variants (`ERA5`, `ERA5ModelLevels`).

Include a reflector so that existing code using `ERA5ARCO` continues to work with a deprecation warning.

### LINKED ISSUE #238: Enable PSFlight to run over a Fleet object
Currently, the `PSFlight` model can only run on a single `Flight`. 

We optimize the `Cocip` model to be able to run on a set of flights (`Fleet`). 

We'd like to enable users to run `PSFlight` standalone with a `Fleet`.

**Alternatives**

Run `Cocip` with a max age of 0.

zebengberg:
Closes #180 
Closes #151
Closes #238 

### Breaking changes

- Remove `attrs["crs"]` usage from `GeoVectorDataset` and child classes (`Flight`, `Fleet`). All spatial data is assumed to be EPSG:4326 (WGS84). This was previously assumed implicitly, but now the `crs` attribute is removed from the `attrs` dictionary.
- Remove deprecated `MetDataset.variables` property in favor of `MetDataset.indexes`.
- Remove `**kwargs` in `MetDataArray` constructor.
- Rename `ARCOERA5` to `ERA5ARCO` for consistency with the `ERA5` and `ERA5ModelLevel` interfaces.

### Features

- Implement `PSFlight.eval` on a `Fleet` source.

## Tests

- [x] CI tests pass

## Reviewer

> Select @ reviewer


tests/unit/test_ps_model.py
```diff
@@ -550,3 +549,22 @@ def test_fill_low_altitude_with_zero_wind(
...
+    assert len(df_flight) == 254
+
+    fl1 = Flight(df_flight.iloc[::2], aircraft_type="A320", flight_id=1)
+    fl2 = Flight(df_flight.iloc[1::2], aircraft_type="A333", flight_id=2)
+    fleet = Fleet.from_seq([fl1, fl2])
+
+    fleet["air_temperature"] = fleet.T_isa()
+
+    ps_model = ps.PSFlight(fill_low_altitude_with_zero_wind=True)
+    out = ps_model.eval(fleet)
```
mlshapiro, line 569: This is a more generic `Fleet` question:

I had a question here recently about how `Fleet` objects handle attrs. Would we be able to have a test here the shows how the attrs show up in "out"?

I would expect to find `aircraft_type` somewhere in the output.

```python
assert "flight_id" in out.data and "aircraft_type" in out.data
```

or is there another place these attrs get stored?

mlshapiro (approved):
This looks good @zebengberg! Thanks for taking the time to do the cleanup.

tests/unit/test_ps_model.py
zebengberg: There's not a great way to handle attrs on `Fleet` instances. The current paradigm is to include global properties in `Fleet.attrs` and individual flight properties in `Fleet.fl_attrs`. I expanded the test a bit to check these out: 5602a03ac229417d3f88c517b7ce55ddc83e4522

We haven't made consistent use of `Fleet.attrs` throughout pycontrails. I think the best practice would be to check for a scalar quantity in `Fleet.attrs` first, then to look within `Fleet.fl_attrs` next ... you can see how this gets complicated tricky.

mlshapiro: oh great! this answers my question. Is there any reason not to use `Fleet.attrs` instead of `Fleet.fl_attrs`? Was it a compatibility or serialization question?

zebengberg: We definitely could have used `Fleet.attrs` instead. We always end up creating a dictionary of the form `flight_id: fl.attrs` for each `fl in fleet`, and we make use of it in various `Fleet` methods, so I attached it as an explicit property on the instance rather than assuming `fleet.attrs` included separate keys for each flight_id.

zebengberg: Okay -- so you're suggesting we overload `Fleet.attrs` to hold `{flight_id: flight_properties}`. I think could be a reasonable way to proceed. I think I avoided this in the past because `attrs["crs"]` was getting in my way.

I think I'll hold off for now, but may do this whenever I iterate again on `Fleet`.

PR #255, opened 2024-10-15
maintenance: replace overrides.overrides with typing.override

zebengberg:
### Internals

- Remove `overrides` dependency. Require `typing-extensions` for python < 3.12.
- Upgrade some type hints for more modern python language features.

## Tests

- [x] CI tests pass

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -170,7 +170,7 @@ autodoc_typehints = "none"
 
 # autodoc options
 autoclass_content = "class"  # only include docstring from Class (not __init__ method)
-autodoc_inherit_docstrings = False
+autodoc_inherit_docstrings = True
 autodoc_default_options = {
     "members": None,  # means yes/true/on
     "undoc-members": None,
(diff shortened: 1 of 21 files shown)


docs/conf.py
```diff
@@ -170,7 +170,7 @@
 
 # autodoc options
 autoclass_content = "class"  # only include docstring from Class (not __init__ method)
-autodoc_inherit_docstrings = False
+autodoc_inherit_docstrings = True
```
zebengberg, line 173: Previously, the `@overrides` decorator copied a docstring to the overriden method if it didn't already have its own. We still want to see the docstring in the built sphinx docs, and this updating this configs seems to achieve that.

mlshapiro (approved):
This all looks good to me, thank you for taking care keeping us up to date with the latest typing conventions.

I used the review opportunity to fix a few docstring issues. There are hundreds more, but it felt good to address a few.

PR #256, opened 2024-10-23
Feature/ps grid mach optimize

trdean1:
## Changes

> Improves computation of mach limits to accept vectorized input/output
> Creates new method `ps_nominal_optimize_mach` which computes the optimal mach number given a set of operating conditions

#### Internals

All functions now use the `newton` optimization method similar to `ps_nominal_grid`.

## Tests

- [ ] CI tests pass

### DIFF
--- a/pycontrails/models/ps_model/ps_grid.py
+++ b/pycontrails/models/ps_model/ps_grid.py
@@ -621,14 +621,7 @@ def ps_nominal_optimize_mach(
         if sin_a is None or cos_a is None:
             msg = "Segment angles must be provide if wind data is specified"
             raise ValueError(msg)
-        headwind = np.sqrt((northward_wind * cos_a) ** 2 + (eastward_wind * sin_a) ** 2)
-        # There's probably a better way to do this?
-        # Compute angle between wind and segment angle and set to (-pi, pi]
-        dth = np.arctan2(cos_a, sin_a) - np.arctan2(eastward_wind, northward_wind)
-        dth = ((dth + np.pi) % (2 * np.pi)) - np.pi
-        # Flip sign of wind if more than pi/2 part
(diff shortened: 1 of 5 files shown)


pycontrails/models/ps_model/ps_grid.py
```diff
@@ -503,3 +503,188 @@ def ps_nominal_grid(
...
+            msg = "Segment angles must be provide if wind data is specified"
+            raise ValueError(msg)
+        headwind = np.sqrt((northward_wind * cos_a) ** 2 + (eastward_wind * sin_a) ** 2)
+        # There's probably a better way to do this?
+        # Compute angle between wind and segment angle and set to (-pi, pi]
+        dth = np.arctan2(cos_a, sin_a) - np.arctan2(eastward_wind, northward_wind)
+        dth = ((dth + np.pi) % (2 * np.pi)) - np.pi
+        # Flip sign of wind if more than pi/2 part
+        sign = 2 * (np.abs(dth) > np.pi / 2) - 1
+        headwind *= sign
```
zebengberg, line 631: So we want to project the wind vector onto the direction of travel of the flight? I think this could be as simple as

```python
headwind = -(northward_wind * cos_a + eastward_wind * sin_a)
```

but I'm getting different values from you when I do this.

In your work, I'm not sure about the first line: `headwind = np.sqrt((northward_wind * cos_a) ** 2 + (eastward_wind * sin_a) ** 2)`.

zebengberg (commented):
Thanks Tom! This looks great. I just left one comment re the headwind formula.

pycontrails/models/ps_model/ps_grid.py
trdean1: Yes you have the right formula...will fix this and the test values. Thanks!

PR #264, opened 2024-11-08
Remove duplicate of potential vorticity parameter

jsmretschnig:
ACCF.met_variables included the same parameter twice: pv (potential_vorticity)

### DIFF
--- a/pycontrails/models/accf.py
+++ b/pycontrails/models/accf.py
@@ -154,7 +154,6 @@ class ACCF(Model):
         RelativeHumidity,
         NorthwardWind,
         EastwardWind,
-        ecmwf.PotentialVorticity,
     )
     sur_variables = (ecmwf.SurfaceSolarDownwardRadiation, ecmwf.TopNetThermalRadiation)
     default_params = ACCFParams


jsmretschnig:
Also: the grib1_id for relative_humidity (r) should be 52 instead of None. See here: [NCEP Grib v1 Code Table](https://www.nco.ncep.noaa.gov/pmb/docs/on388/table2.html)
Can't find it in the code though. The unit seems correct (%), but it's not consistent with ACCF.met_variables, there it says 1.

To reproduce the issue: try to download ERA5 data with ACCF.met_variables as parameter

zebengberg:
> Also: the grib1_id for relative_humidity (r) should be 52 instead of None. See here: [NCEP Grib v1 Code Table](https://www.nco.ncep.noaa.gov/pmb/docs/on388/table2.html) Can't find it in the code though. The unit seems correct (%), but it's not consistent with ACCF.met_variables, there it says 1.
> 
> To reproduce the issue: try to download ERA5 data with ACCF.met_variables as parameter

Thanks @jsmretschnig ! This is certainly confusing.

pycontrails has two different `RelativeHumidity` variables:

- [A "core" RH with `grib1_id=52`](https://github.com/contrailcirrus/pycontrails/blob/1b3b9b21303fecf774272c893079f423b33fe07a/pycontrails/core/met_var.py#L179). This gives the humidity as a proportion.
- [An "ecmwf" flavored RH with `grid1_id=None`](https://github.com/contrailcirrus/pycontrails/blob/1b3b9b21303fecf774272c893079f423b33fe07a/pycontrails/datalib/ecmwf/variables.py#L104). This gives the humidity as a percentage.

Can you confirm that ACCFs expect the first flavor of RH? If so, I'll figure out how to fix this.

jsmretschnig:
So apparently CLIMaCCF is fine with both options: https://github.com/dlr-pa/climaccf/blob/907c5f0cc84db8be675337d6a213c8706b024409/climaccf/weather_store.py#L277

And uses this parameter https://codes.ecmwf.int/grib/param-db/157

PR #269, opened 2024-11-14
add validation handler and deps

nickmasson:
#### Features
Adds a `ValidateTrajectoryHandler` to `pycontrails.datalib.spire`.

The handler takes a pandas dataframe, where each row represents a single flight instance trajectory waypoint.

Certain columns are required (`ValidateTrajectoryHandler.SCHEMA`) beyond what is required in e.g. a `Flight` instance (`pycontrails.core.flight`), as those additional columns are used in the validation ruleset that evaluates the trajectory.

### TODO
- [ ] mypy wrestling
- [ ] update changelog
- [ ] update release notes
- [ ] add tutorial/cheat-sheet (proposed location: https://py.contrails.org/flight.html)
- [ ] (?) additional consolidation and structuring of existing methods in `spire.py` (could also be a follow-up release, as those changes will be backward incompatible)

## Tests

- [ ] CI tests pass


pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
 
     # reset original index and return
     return resampled.reset_index(drop=True)
+
+
+class ValidateTrajectoryHandler:
+    """
+    Evaluates a trajectory and identifies if it violates any verification rules.
+
+    <LINK HERE TO HOSTED REFERENCE EXAMPLE(S)>.
```
nickmasson, line 763: TODO

pycontrails/datalib/exceptions.py
```diff
@@ -0,0 +1,58 @@
...
+
+class FlightTooSlowError(Exception):
+    """Trajectory has period(s) of unrealistically slow speed."""
+
+
+class FlightTooFastError(Exception):
+    """Trajectory has period(s) of unrealistically high speed."""
+
+
+class RocdError(Exception):
```
zebengberg, line 45: Per PEP8, this should be `ROCDError`

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+            ],
+        )
+        self._df = self._df.assign(
+            ground_speed_m_s=self._df["elapsed_distance_m"]
+            .divide(self._df["elapsed_seconds"])
+            .replace(np.inf, np.nan)
+        )
+        self._df = self._df.assign(
+            rocd_fps=[self._rolling_rocd_fps(window) for window in self._df.rolling(window=2)]
+        )
```
zebengberg, line 1060: My main piece of feedback for this PR: there are a number of `pycontrails.Flight` methods and other helper functions that already do these computations in a vectorized manner. Some of them may need slight tweaking to match exactly what you're doing here, but it shouldn't be too hard. Is there a reason to avoid them?

If you have a good reason to avoid the existing pycontrails tooling (and there definitely are some good reasons -- some of the existing pycontrails function have opinions / conventions baked into them), all of these calculations can still be vectorized. For example, your calculation for `elapsed_seconds` is essentially:

```
elapsed_seconds = self._df["timestamp"].diff().dt.total_seconds().astype("Int64")
```

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+
+        if len(self._df["arrival_airport_icao"].value_counts()) > 1:
```
zebengberg, line 1062: `len(self._df["arrival_airport_icao"].value_counts())` -> `self._df["arrival_airport_icao"].nunique()`

(here and below)

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+            "airline_iata",
+            "departure_airport_icao",
+            "departure_scheduled_time",
+            "arrival_airport_icao",
+            "arrival_scheduled_time",
+        ]
+
+        violations = []
+        for k in invariant_fields:
+            unique_vals = list(self._df[k].value_counts().index)
```
zebengberg, line 1139: This can also be changed to use `pd.DataFrame.nunique`

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+        # we do this because some of the following checks assume no invariant field violations,
+        #   or timestamp dupes
+        if len(all_violations) > 0:
+            return all_violations
+
+        # Checks; Round 3
+        self._calculate_additional_fields()
+
+        flight_length_check: None | FlightTooShortError | FlightTooLongError
+        flight_length_check = self._is_valid_flight_length()
```
zebengberg, line 1392: You should be able to delete the line `flight_length_check: None | FlightTooShortError | FlightTooLongError`  -- the type hint is already provided on the method `self._is_valid_flight_length()`.

zebengberg (commented):
Thanks @nickmasson! I made a ton of comments. Let me know if I can better explain anything!

I'm also happy to do some work here to better use some of the existing pycontrails tooling.

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+        "arrival_scheduled_time": pdtypes.is_datetime64_any_dtype,
+        "ingestion_time": pdtypes.is_datetime64_any_dtype,
+        "timestamp": pdtypes.is_datetime64_any_dtype,
+        "latitude": pdtypes.is_numeric_dtype,
+        "longitude": pdtypes.is_numeric_dtype,
+        "collection_type": pdtypes.is_string_dtype,
+        "altitude_baro": pdtypes.is_numeric_dtype,
+    }
+
+    airports_db = airports.global_airport_database()
```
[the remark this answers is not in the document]
nickmasson: 👍 
The way the handler is designed, an implementer shouldn't be reinstantiating an instance of the class multiple times in a given runtime... so this should (ideally) be loaded only once per program. But I don't see a reason using the functools cacher would hurt, so happy to add (we could also move this to a module-level singleton, so it is similarly loaded only once per program, with more confidence than if it were a class obj).

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+
+
+class ValidateTrajectoryHandler:
+    """
+    Evaluates a trajectory and identifies if it violates any verification rules.
+
+    <LINK HERE TO HOSTED REFERENCE EXAMPLE(S)>.
+    """
+
+    CRUISE_ROCD_THRESHOLD_FPS = 4.2  # 4.2 ft/sec ~= 250 ft/min
```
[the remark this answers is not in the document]
nickmasson: yep, absolutely.  It is OK if I keep it as FPS within the boundaries of the handler? It will just save having to refactor broadly.

What I'll do is pull this value in from `constants.py`, but still do it as a class variable, so that an implementer can monkey-patch the handler if they want.

pycontrails/datalib/spire.py
nickmasson: Great. No reason to have avoided methods that have parity of behavior using a vectorized approach (other than my fairly shallow experience w. pd and np).  I'd like to avoid referencing existing pycontrails methods that have domain logic/assumptions, but otherwise welcome refactor of these to avoid `pd.rolling()` (I assume that will help w/ perf notably).

@zebengberg would you mind helping with that refactor? Feel free to push changes directly to this branch, or throw up small PRs pointed to this branch with the proposed refactor.

pycontrails/datalib/spire.py
nickmasson: Agreed w.r.t type hinting and mypy etc...

The hinting here was motivated as a hint in the spirit of how a developer might use the handler.

Given that the handler returns a list of one or more typed violations if one or more violations are encountered, my expectation is that a developer would likely use this method definitions in drafting their approach to managing the handler response. It seems to me like having a hint of the possible objects aggregated & returned into `all_violations` will help a developer avoid drilling up and down into all the methods.  If these were removed, then someone looking to understand the full scope of what the handler is returned would have to traverse a lot more src code.

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+    @staticmethod
+    def _calc_distance_m(lat_0, lon_0, alt_ft_0, lat_f, lon_f, alt_ft_f) -> float:
+        """Calculate great circle distance between two lat/lon/alt coordinates."""
+        dist_m = math.sqrt(
+            (0.3048 * (alt_ft_f - alt_ft_0)) ** 2
+            + geo.haversine(
+                lons1=np.array(lon_f),
+                lats1=np.array(lat_f),
+                lons0=np.array(lon_0),
+                lats0=np.array(lat_0),
```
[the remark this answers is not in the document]
nickmasson: 👍 I recall having it codified that way first but changed to passing arrays for some reason... I'll revisit and simplify if I was ill informed.

pycontrails/datalib/spire.py
```diff
@@ -737,3 +754,663 @@ def _downsample_flight(
...
+        if len(matches) == 0:
+            return np.nan, np.nan, np.nan
+        if len(matches) > 1:
+            raise ValueError(
+                f"found multiple matches for aiport icao {airport_icao} " f"in airports database."
+            )
+
+        lat = matches.iloc[0]["latitude"]
+        lon = matches.iloc[0]["longitude"]
+        alt_ft = matches.iloc[0]["elevation_ft"]
```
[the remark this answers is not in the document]
nickmasson: I'm seeing other places where row/col selection could/should be switched. Many of those are in the methods applied in the rolling computations, which may evaporate with refactor to use vectorized methods. I'll leave them for now, and if any persist after the refactor, they can be changed.

pycontrails/datalib/spire.py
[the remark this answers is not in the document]
nickmasson: `geo.haversine` expects the inputs as `ArrayLike`... but the implementation supports passing `float`s. I'll change to just pass floats here.  Should the typehint of geo.haversine be updated?

pycontrails/datalib/spire.py
```diff
@@ -2,738 +2,583 @@
...
-    "aircraft_type_icao": str,
-    "aircraft_type_name": str,
-    "airline_iata": str,
-    "airline_name": str,
-    "departure_utc_offset": str,
-    "departure_scheduled_time": str,
-}
-
-
-def clean(messages: pd.DataFrame) -> pd.DataFrame:
```
zebengberg, line 40: @mlshapiro I removed all of the legacy code that was previously in this spire.py module in favor of @nickmasson's update. Let me know if you'd want me to retain anything. The implementations are somewhat different -- it looks like your previous work did a bunch of filtering, whereas the new implementation flags violations.

PR #270, opened 2024-11-15
Feature/load-factor-improvements

roger-teoh:
Closes #XX

## Changes

- Adds function to estimate aircraft (passenger/cargo) load factor based on historical monthly and regional load factors provided by IATA. This improves upon the default load factor assumption. 
- Historical load factor databases will be continuously updated as new data is released. 

#### Internals

- Update default load factor from 70% to 83% to be consistent with historical data

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer


pycontrails/physics/jet.py
```diff
@@ -8,15 +8,20 @@
...
 import numpy.typing as npt
 
 from pycontrails.core import flight
 from pycontrails.physics import constants, units
 from pycontrails.utils.types import ArrayOrFloat, ArrayScalarLike
 
 logger = logging.getLogger(__name__)
+_path_to_static = pathlib.Path(__file__).parent / "static"
+PLF_PATH = _path_to_static / "iata-passenger-load-factors-20241115.csv"
+CLF_PATH = _path_to_static / "iata-cargo-load-factors-20241115.csv"
```
zebengberg, line 24: Can you confirm that both of these files are fine to share publicly?

zebengberg:
@roger-teoh -- I'll make a few small commits to this branch to avoiding reading the csvs whenever the jet module is imported.

pycontrails/physics/jet.py
roger-teoh: Yes, I have created these csv’s from reports that are publicly available and published by IATA

PR #282, opened 2024-12-09
fix CocipGrid met downselection in time

zebengberg:
#### Fixes

- Fix the `CocipGrid` met downselection process to accommodate cases where `dt_integration` is as large as the time step of the met data. Previously, the output of `CocipGrid(met=met, rad=rad, dt_integration="1 hour")` was zero everywhere when the `met` and `rad` data had a time step of 1 hour.

## Tests

- [x] CI tests pass

### DIFF
--- a/tests/unit/test_cocip_grid.py
+++ b/tests/unit/test_cocip_grid.py
@@ -324,7 +324,7 @@ def test_grid_survival_fraction(instance_params: dict[str, Any], source: MetData
         ),
         (
             np.datetime64("2019-01-01T12:00"),
-            [np.datetime64("2019-01-01T11:00"), np.datetime64("2019-01-01T12:00")],
+            [np.datetime64("2019-01-01T12:00")],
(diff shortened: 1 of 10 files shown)


tests/unit/test_cocip_grid.py
```diff
@@ -324,7 +324,7 @@ def test_grid_survival_fraction(instance_params: dict[str, Any], source: MetData
         ),
```
thabbott, line 324: Probably worth adding a test for a `CocipGrid(met=met, rad=rad, dt_integration="1 hour")` case that would have failed before these changes.

thabbott (approved):
Looks good to me!

We talked about basically everything in person--just a couple of minor comments.

tests/unit/test_cocip_grid.py
zebengberg: 266176f79e9fa1bd2de325b55670d47f3f101513

PR #287, opened 2025-01-16
include additional variables in the DryAdvection output

zebengberg:
### Features

- Add two new parameters to the `DryAdvection` model.
  - If the `verbose_outputs` parameter is enabled, additional wind-shear data is included in the output.
  - If the `include_source_in_output` parameter is enabled, the source data with any of the intermediate artifacts (e.g., interpolated met data, wind-shear data, etc.) is included in the output.

  Both parameters are disabled by default.

## Tests

- [x] CI tests pass

### DIFF
--- a/pycontrails/models/dry_advection.py
+++ b/pycontrails/models/dry_advection.py
@@ -63,6 +63,15 @@ class DryAdvectionParams(models.AdvectionBuffers):
     # If None, only pointwise advection is simulated without wind shear effects.
     azimuth: float | None = 0.0
 
+    #: Add additional intermediate variables to the output vector.
+    #: This includes interpolated met variables and wind-shear-derived geometry.
+    verbose_outputs: bool = False
+
+    #: Whether to include ``source`` points in the output vector. Enabling allows
+    #: the user to view additional data (e.g., interpolated met variables) for
+    #: source points as well as evolved points.
+    include_source_in_output: bool = False
+
 
 class DryAdvection(models.Model):
     """Simulate "dry advection" of an emissions plume with an elliptical cross section.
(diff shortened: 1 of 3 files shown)


pycontrails/models/dry_advection.py
```diff
@@ -154,42 +163,51 @@ def eval(self, source: GeoVectorDataset | None = None, **params: Any) -> GeoVect
...
-                vector,
+                vector1,
                 t,
                 sedimentation_rate=sedimentation_rate,
                 dz_m=dz_m,
                 max_depth=max_depth,
+                verbose_outputs=verbose_outputs,
                 **interp_kwargs,
             )
+            evolved.append(vector1)
```
thabbott, line 195: It took me a while to convince myself I understand why you're appending `vector1` but not until after the call to `_evolve_one_step`. Is this so you include source points in the output (why you append `vector1`) with interpolated properties attached (why you append after `_evolve_one_step`)?

thabbott (approved):
This looks good to me! Just one small comment on a part of the changes that confused me a first. Might be worth an explanatory comment in the code if you also think there are subtleties that are worth flagging.

pycontrails/models/dry_advection.py
zebengberg: Sorry for not providing more context!

Yes, your thinking is absolutely correct. If the source points are `waypoints_0`, and each subsequent advection is `waypoints_1`, `waypoints_2`, ... `waypoints_n`, then previously we were collecting and concatenating `[waypoints_1, waypoints_2, ..., waypoints_n]`. With this update, the list is now `[waypoints_0, waypoints_1, waypoints_2, ..., waypoints_n]`.

The `_evolve_one_step` method is full of side effects (ie, mutating the `vector` parameter in place), so it doesn't actually matter if this `evolved.append` occurs before or after the call to `_evolve_one_step`.

PR #291, opened 2025-01-22
Helper methods for generating lists of required met variables

thabbott:
Changes in this PR allow code snippets like
```python
variables = (
   v[0] if isinstance(v, tuple) else v for v in Cocip.met_variables
)
met.standardize_variables(variables)
```
(frequently used to standardize ECMWF met variables, but broken as of the latest pycontrails release) to be replaced with
```python
met.standardize_variables(Cocip.ecmwf_met_variables)
```

#### Features

- Add helper classmethods to `Model`, `Cocip`, and `CocipGrid` for generating lists of required variables from specific data sources.

## Tests

- [ ] CI tests pass

## Reviewer

> @zebengberg

### DIFF
--- a/pycontrails/core/models.py
+++ b/pycontrails/core/models.py
@@ -289,7 +289,8 @@ class Model(ABC):
         tuple[MetVariable]
             List of model-agnostic variants of required variables
         """
-        return tuple(_find_match(required, set(MET_VARIABLES)) for required in cls.met_variables)
+        available = set(MET_VARIABLES)
+        return tuple(_find_match(required, available) for required in cls.met_variables)
 
     @classmethod
     def ecmwf_met_variables(cls) -> tuple[MetVariable, ...]:
@@ -300,7 +301,8 @@ class Model(ABC):
         tuple[MetVariable]
             List of ECMWF-specific variants of required variables
         """
-        return tuple(_find_match(required, set(ECMWF_VARIABLES)) for required in cls.met_variables)
+        available = set(ECMWF_VARIABLES)
+        return tuple(_find_match(required, available) for required in cls.met_variables)
 
     @classmethod
     def gfs_met_variables(cls) -> tuple[MetVariable, ...]:
@@ -311,7 +313,8 @@ class Model(ABC):
         tuple[MetVariable]
             List of GFS-specific variants of required variables
         """
-        return tuple(_find_match(required, set(GFS_VARIABLES)) for required in cls.met_variables)
+        available = set(GFS_VARIABLES)
+        return tuple(_find_match(required, available) for required in cls.met_variables)
 
     def _verify_met(self) -> None:
         """Verify integrity of :attr:`met`.
--- a/pycontrails/core/models.py
+++ b/pycontrails/core/models.py
@@ -289,7 +289,7 @@ class Model(ABC):
         tuple[MetVariable]
             List of model-agnostic variants of required variables
         """
-        return tuple(_find_match(required, MET_VARIABLES) for required in cls.met_variables)
+        return tuple(_find_match(required, set(MET_VARIABLES)) for required in cls.met_variables)
 
     @classmethod
     def ecmwf_met_variables(cls) -> tuple[MetVariable, ...]:
@@ -300,7 +300,7 @@ class Model(ABC):
         tuple[MetVariable]
             List of ECMWF-specific variants of required variables
         """
-        return tuple(_find_match(required, ECMWF_VARIABLES) for required in cls.met_variables)
+        return tuple(_find_match(required, set(ECMWF_VARIABLES)) for required in cls.met_variables)
 
     @classmethod
     def gfs_met_variables(cls) -> tuple[MetVariable, ...]:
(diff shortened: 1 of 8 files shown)


pycontrails/core/models.py
```diff
@@ -276,6 +276,39 @@ def hash(self) -> str:
...
+    @classmethod
+    def ecmwf_met_variables(cls) -> tuple[MetVariable, ...]:
+        """Return an ECMWF-specific list of required meteorology variables.
+
+        Returns
+        -------
+        tuple[MetVariable]
+            List of ECMWF-specific variants of required variables
+        """
+        return tuple(v[1] if isinstance(v, tuple) else v for v in cls.met_variables)
```
zebengberg, line 299: Do you think an implementation like this would be more maintainable? What you have is also good, just wondering if we should avoid relying on the ordering of a tuple. The implementation below requires updating `ECMWF_VARIABLES` when we add new stuff (I don't know if we've been doing this).

```python
from pycontrails.core.met_var import MetVariable
from pycontrails.datalib.ecmwf import ECMWF_VARIABLES

def find_match(vars, available):
    if isinstance(vars, MetVariable):
        return vars
    for var in vars:
        if var in available:
            return var
    raise ...


def ecmwf_met_variables():
    available = set(ecmwf.ECMWF_VARIABLES)
    return tuple(find_match(vars, available) for vars in Cocip.met_variables)
```

zebengberg (approved):

pycontrails/core/models.py
thabbott: I don't have a super strong opinion.... I went with what seemed like the simpler implementation (rather than your suggestion, which is probably closer to maintenance-free) mostly because the point of adding a "generic" option was to avoid having to touch `Cocip.met_variables` (and everything downstream of it) in the future.

But perhaps I am being too optimistic that this means we won't end up annoyed about relying on the ordering of a tuple. I'll think about it a little more before I merge.

thabbott: I've come around to your point of view--see https://github.com/contrailcirrus/pycontrails/pull/291/commits/4827ac388fc33536adc8ec59aa01a3f0e27f9b60. Will wait until tomorrow to merge, but up to you whether you want to take another look. (It's a pretty low-stakes change.)

pycontrails/models/cocipgrid/cocip_grid.py
```diff
@@ -435,6 +436,45 @@ def _bundle_results(
...
+
+        Returns
+        -------
+        tuple[MetVariable]
+            List of GFS-specific variants of required variables
+        """
+        return tuple(
+            models._find_match(required, gfs.GFS_VARIABLES) for required in cls.rad_variables
+        )
+
```
zebengberg, line 477: You might want to replace all this with the methods you've already included on `Cocip` (for example, that's already how `CocipGrid.met_variables` is defined).

So something like this:

```python
generic_rad_variables = cocip.Cocip.generic_rad_variables
ecmwf_rad_variables = cocip.Cocip.ecmwf_rad_variables
gfs_rad_variables = cocip.Cocip.gfs_rad_variables
```

pycontrails/core/models.py
```diff
@@ -805,6 +842,42 @@ def _interp_grid_to_grid(
...
+    Raises
+    ------
+    KeyError
+        Raised if not match is found
+    """
+    if isinstance(required, MetVariable):
+        return required
+
+    for var in required:
+        if var in available:
```
zebengberg, line 872: Small nit: I think you should either require `available` to be a `set[MetVariable]`, or explicitly create a `set(available)` here to make this `var in available` check constant time. For example:

```python
available_set = set(available)
for var in required:
   if var in available_set:
      ...
```

Another similar option is

```python
matched = set(available).intersection(required)
if len(matched) == ...
```

zebengberg (approved):

pycontrails/models/cocipgrid/cocip_grid.py
thabbott: https://github.com/contrailcirrus/pycontrails/pull/291/commits/0a428809bf43f3a81d58e03b1dc7e01619d6b175

pycontrails/core/models.py
thabbott: https://github.com/contrailcirrus/pycontrails/pull/291/commits/754b268f486a3133392effbbf2e26997956ae7e5

pycontrails/core/models.py
```diff
@@ -276,6 +280,39 @@ def hash(self) -> str:
...
+    @classmethod
+    def generic_met_variables(cls) -> tuple[MetVariable, ...]:
+        """Return a model-agnostic list of required meteorology variables.
+
+        Returns
+        -------
+        tuple[MetVariable]
+            List of model-agnostic variants of required variables
+        """
+        return tuple(_find_match(required, set(MET_VARIABLES)) for required in cls.met_variables)
```
zebengberg, line 292: Here (and in the other five places), I'd suggest tweaking to:

```python
available = set(MET_VARIABLES)
return tuple(_find_match(required, available) for required in cls.met_variables)
```

This avoids creating the set repeatedly.

thabbott: https://github.com/contrailcirrus/pycontrails/pull/291/commits/ee2a74d93dbc948c3863530ce051abc42b16ce9e

PR #302, opened 2025-03-04
Add model level support to Teoh humidity scaling

### LINKED ISSUE #301: ExponentialBoostHumidityCorrection coefficients
**Description**

The coefficients used in ExponentialBoostHumidityScaling() are calibrated using ERA5 reanalysis pressure level data. Therefore, the coefficients are not valid when used on model levels. 

The function needs to identify if the ERA5 model level or pressure level data is used, then use the right coefficients. The model level coefficients should be as below, while pressure level coefficients remains the same.

a1=0.026302083473778933,
a2=0.9651041666272301,
a3=2.250107435038734,
a4=36.549393263544914,

b1=0.489062500000000095,
b2=2.210937499999999,
b3=4.182743010484767,
b4=17.53378893219587,

**Details**

- Version: [e.g. 0.54.8]
- Module: humidity_scaling.py
- OS: [e.g. ubuntu, windows]
- Reporter:

thabbott:
Closes #301 

#### Features

- Add `ExponentialBoostLatitudeCorrectionHumidityScaling` calibrated for model-level ERA5 data.

Creating an instance of `ExponentialBoostLatitudeCorrectionHumidityScaling` without explicitly selecting model- or pressure-level coefficients will use pressure-level coefficients by default (which preserves current behavior) but warn users that a future release will default to using model-level coefficients. This mirrors the behavior of the `HistogramMatching` humidity scaling, which also includes separate pressure- and model-level calibrations.

## Tests

- [ ] CI tests pass

## Reviewer

> @zebengberg, @roger-teoh


pycontrails/models/humidity_scaling/humidity_scaling.py
```diff
@@ -321,7 +321,22 @@ class _SigmoidCoefficients:
 
 
 @functools.cache
-def _load_sigmoid_coef(q_method: str | None) -> _SigmoidCoefficients:
+def _load_sigmoid_coef(q_method: str | None, level_type: str | None) -> _SigmoidCoefficients:
```
zebengberg, line 324: Should `level_type` be a `str` (not `str | None`)?

pycontrails/models/humidity_scaling/humidity_scaling.py
```diff
@@ -406,12 +429,30 @@ class ExponentialBoostLatitudeCorrectionHumidityScaling(HumidityScaling):
...
     scaler_specific_keys = ("latitude",)
 
+    def __init__(
+        self,
+        met: MetDataset | None = None,
+        params: dict[str, Any] | None = None,
+        **params_kwargs: Any,
+    ):
+        if (params is None or "level_type" not in params) and (
+            params_kwargs is None or "level_type" not in params_kwargs
```
zebengberg, line 442: I don't think the `param_kwargs is None` situation is possible here (correct me if I'm wrong).

pycontrails/models/humidity_scaling/humidity_scaling.py
```diff
@@ -321,7 +321,22 @@ class _SigmoidCoefficients:
...
+        return _SigmoidCoefficients(
+            a1=0.026302083473778933,
+            a2=0.9651041666272301,
+            a3=2.250107435038734,
+            a4=36.549393263544914,
+            b1=0.489062500000000095,
+            b2=2.210937499999999,
+            b3=4.182743010484767,
+            b4=17.53378893219587,
+        )
```
zebengberg, line 335: Did you compute these with linear interpolation? If so, should we mandate `q_method=None` (for example, raise a ValueError if `q_method` is not None)?

zebengberg (approved):
Thanks Tristan!

Please update the CHANGELOG as well.

pycontrails/models/humidity_scaling/humidity_scaling.py
thabbott: No, I think you're right: 9ebbe7e

pycontrails/models/humidity_scaling/humidity_scaling.py
thabbott: @roger-teoh

thabbott:
Thanks, Zeb!

I'll wait for input from @roger-teoh on one of your comments and then merge.

pycontrails/models/humidity_scaling/humidity_scaling.py
zebengberg: 96a1c7c7c93c7d12c2b8dc540db3f6cb4a26b95e

PR #308, opened 2025-04-29
Feature/goes parallax

zebengberg:
## Changes

- Include new `goes.correct_parallax` function in the `GOES` datalib to correct for parallax artifacts in GOES imagery. This function can be used to better superimpose flight and contrail data over GOES imagery.

## Tests

- [x] CI tests pass

## Reviewer

> @thabbott

### DIFF
--- a/pycontrails/datalib/goes.py
+++ b/pycontrails/datalib/goes.py
@@ -780,3 +780,108 @@ def _clip_and_scale(
         Clipped and scaled array.
     """
     return (arr.clip(low, high) - low) / (high - low)
+
+
+def parallax_correct(
+    longitude: npt.NDArray[np.floating],
+    latitude: npt.NDArray[np.floating],
+    altitude: npt.NDArray[np.floating],
+    goes_da: xr.DataArray,
+) -> tuple[npt.NDArray[np.floating], npt.NDArray[np.floating]]:
+    """Apply parallax correction to WGS84 geodetic coordinates based on satellite perspective.
+
+    This function considers the ray from the satellite to the points of interest and finds
+    the intersection of this ray with the WGS84 ellipsoid. The intersection point is then
+    returned as the corrected longitude and latitude coordinates.
+
+    This function requires the :mod:`pyproj` package to be installed.
+
+    Parameters
+    ----------
+    longitude : npt.NDArray[np.floating]
+        A 1D array of longitudes in degrees.
+    latitude : npt.NDArray[np.floating]
+        A 1D array of latitudes in degrees.
+    altitude : npt.NDArray[np.floating]
+        A 1D array of altitudes in meters.
+    goes_da : xr.DataArray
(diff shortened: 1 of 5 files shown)


pycontrails/datalib/goes.py
```diff
@@ -770,3 +784,122 @@ def _clip_and_scale(
...
+    return inv_transformer.transform(*intersection)[:2]  # final coord is (close to) 0
+
+
+def _intersection_with_ellipsoid(
+    p0: npt.NDArray[np.floating],
+    p1: npt.NDArray[np.floating],
+    a: float,
+    b: float,
+) -> npt.NDArray[np.floating]:
+    """Find the intersection of a line with the surface of the WGS84 ellipsoid."""
```
thabbott, line 873:
```suggestion
    """Find the intersection of a line with the surface of an ellipsoid."""
```
(This function isn't WGS84-specific since you're passing in `a` and `b`.)

pycontrails/datalib/goes.py
```diff
@@ -770,3 +784,122 @@ def _clip_and_scale(
...
+
+    # Calculate the two possible solutions for t
+    t0 = (-B + sqrtD) / (2.0 * A)
+    t1 = (-B - sqrtD) / (2.0 * A)
+
+    # Calculate the intersection points
+    intersection0 = p0 + t0 * v
+    intersection1 = p0 + t1 * v
+
+    # CRUDE: Pick the intersection point that is closer to the satellite
```
thabbott, line 902: I'm not sure this is the most intuitive choice. If the plane (`p1`) is on the opposite side of the planet I'd expect either to get a NaN or the farther intersection point, not the closer one.

tests/unit/test_goes.py
```diff
@@ -139,3 +139,45 @@ def test_goes_get_with_cache(t: str, cachestore: DiskCacheStore) -> None:
 
     t_str = pd.Timestamp(t).strftime("%Y%m%d%H%M")
     assert cachestore.listdir() == [f"M2_{t_str}_C02.nc"]
+
+
+def test_goes_parallax_correct():
```
thabbott, line 144: Suggest adding a test for a plane on the opposite side of the planet.

pycontrails/datalib/goes.py
zebengberg: Good point, thanks for thinking this through. I could simply pick the intersection point closer to the aircraft here. I'm inclined to do that and just issue a warning when we're an the opposite side of the planet. I'll try something.

tests/unit/test_goes.py
zebengberg: @thabbott Added a test for this case! 6810095bbd79e6fc6346493c3be699f54f0d48fc

pycontrails/datalib/goes.py
thabbott: Your updates look good to me!

PR #309, opened 2025-05-01
Update `AircraftPerformance` base class

thabbott:
Addresses some sharp corners that @nickmasson encountered while implementing a new `AircraftPerformance` subclass.

`BADAFlight` should also be updated once these changes are finalized and merged.

### Breaking changes

- Air temperature is now an optional met variable in `PSFlight`. Code that uses `PSFlight.met_variables` may need to be updated.
- The `ensure_true_airspeed_on_source` method of the `AircraftPerformance` class now operates fully in-place and no longer returns a value. Code that relies on a return value may need to be updated.

### Internals

- Define `default_parameters`, `met_variables`, and `optional_met_variables` in the `AircraftPerformance` abstract base class.
- Remove the definitions of `met_variables` and `optional_met_variables` from `PSFlight`. These are now inherited from `AircraftPerformance`.

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg


pycontrails/core/aircraft_performance.py
```diff
@@ -9,9 +9,9 @@
 from typing import Any, Generic, NoReturn, overload
 
 if sys.version_info >= (3, 12):
-    from typing import override
+    pass
```
thabbott, line 12: @zebengberg: I'm pretty sure I didn't edit these lines myself, so not sure why they've changed... I assume I should revert?

zebengberg: Yes please! Perhaps you editor tried to do some magic. I don't think this is a ruff formatting thing.

pycontrails/core/aircraft_performance.py
```diff
@@ -491,25 +479,59 @@ def calculate_aircraft_performance(
...
+                "Cannot compute air temperature without providing met data that includes an "
+                "'air_temperature' variable. Either include met data with 'air_temperature' "
+                "in the model constructor, define 'air_temperature' data on the flight, or set "
+                "'fill_low_altitude_with_isa_temperature' to True."
+            )
+            raise ValueError(msg)
+
+        self.source["air_temperature"] = interpolate_met(
+            self.met, self.source, "air_temperature", **self.interp_kwargs
+        )
```
zebengberg, line 512: The `interpolate_met` function might already set the field on `self.source`. If so, you could get rid of the `self.source["air_temperature"] = ` part. (I also thought we'd get a warning if we try to overwrite some field ... but I didn't see any warnings come out of the pytest suite. Perhaps we never actually test this line of code?)

pycontrails/core/aircraft_performance.py
```diff
@@ -491,25 +479,59 @@ def calculate_aircraft_performance(
...
-            using :meth:`Flight.segment_true_airspeed`.
+        This function operates in-place. If ``true_airspeed`` is not already present
+        on :attr:`source`, it is calculated using :meth:`Flight.segment_true_airspeed`.
         """
         tas = self.source.get("true_airspeed")
         fill_with_groundspeed = self.params["fill_low_altitude_with_zero_wind"]
 
         if tas is not None:
             if not fill_with_groundspeed:
-                return tas
```
zebengberg, line 509: This refactor looks good. Originally I had some reason for returning the true airspeed, but it doesn't appear to be relevant anymore.

zebengberg (approved):
Thanks Tristan!

If you haven't done this already, I think you should run the pycontrails-bada test suite against these changes.

pycontrails/core/aircraft_performance.py
thabbott: [`test_fill_low_altitude_with_isa_temperature`](https://github.com/contrailcirrus/pycontrails/blob/ec687aa8b3f3da269858cd84c5b8619813f2462b/tests/unit/test_ps_model.py#L585) should run this line.

I think the reason we're not getting a warning is this because is because we don't raise a warning when we overwrite a field with itself. (i.e., `self.source[k] = v` won't produce a warning if `self.source[k]` and `v` refer to the same array).

Still, you're right that we don't need this assignment! I'll get rid of it.

pycontrails/core/aircraft_performance.py
thabbott: c819224

zebengberg (approved):

pycontrails/ext/synthetic_flight.py
```diff
@@ -207,7 +207,7 @@ def __call__(self, timestep: np.timedelta64 | None = None) -> Flight:
 
     def _id(self) -> int:
         """Get random flight ID."""
-        return self.rng.integers(100_000, 999_999).item()
+        return self.rng.integers(100_000, 999_999)
```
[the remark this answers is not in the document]
thabbott: This change "fixed" an error I was getting because my mypy was outdated. I also need the `.item()` with an up-to-date mypy, so it's back (https://github.com/contrailcirrus/pycontrails/pull/309/commits/b02be7e6ec9dfb13f04ecb7bdfbc141e0cd28f91)

PR #318, opened 2025-06-04
fix goes caching for custom bucket

zebengberg:
### Fixes

- Fix the `GOES` interface caching mechanism when a custom `goes_bucket` is provided. This allows users to cache both GOES-16 and GOES-18 data without conflicts.

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/pycontrails/datalib/goes.py
+++ b/pycontrails/datalib/goes.py
@@ -453,7 +453,10 @@ class GOES:
 
     def __repr__(self) -> str:
         """Return string representation."""
-        return f"GOES(region='{self.region}', channels={sorted(self.channels)})"
+        return (
+            f"GOES(region={self.region}, channels={sorted(self.channels)}, "
+            f"bucket={self.goes_bucket})"
+        )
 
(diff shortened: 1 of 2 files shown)


pycontrails/datalib/goes.py
```diff
@@ -453,7 +453,10 @@ def __init__(
 
     def __repr__(self) -> str:
         """Return string representation."""
-        return f"GOES(region='{self.region}', channels={sorted(self.channels)})"
+        return (
+            f"GOES(region={self.region}, channels={sorted(self.channels)}, "
+            f"bucket={self.goes_bucket})"
```
thabbott: Is this what you want when `self.goes_bucket` is `None`?

thabbott (approved):
One small comment--looks fine to me!

pycontrails/datalib/goes.py
zebengberg: Yeah -- I briefly considered this. I think it's okay to keep as is -- when `None`, the `goes_bucket` is defined when the function `gcs_goes_path` is called, and it'd be hard to back this out to the `GOES` interface. I will change `bucket` -> `goes_bucket`.

PR #320, opened 2025-06-09
Replace newlines with spaces instead of empty strings

trobarge26:
#### Fixes
Corrects an issue that when parsing a flight plan string, if the route contains newlines without spaces, it concatenates multiple route elements, causing Skyvector to ignore those elements. This replaces the newlines with a space instead of an empty string. 

## Tests

- [ ] CI tests pass

## Reviewer

> @trdean1

### DIFF
--- a/pycontrails/core/flightplan.py
+++ b/pycontrails/core/flightplan.py
@@ -97,8 +97,8 @@ def parse_atc_plan(atc_plan: str) -> dict[str, str]:
     --------
     :func:`to_atc_plan`
     """
-    atc_plan = atc_plan.replace("\r", "")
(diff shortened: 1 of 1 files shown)


zebengberg:
@trobarge26 I think some of your changes need formatting. See the [dev instructions here](https://py.contrails.org/develop.html). If I'm understanding correct your work correctly, everything besides your two changes replace `""` with `" "` should be reverted.

trobarge26:
Thanks Zeb. I think an auto-formatting extension I have installed made changes I wasn't intending - the only changes should have been to those two lines. I'll try again and resubmit.

PR #321, opened 2025-06-12
Update flightplan.py

trobarge26:
When parsing a flightplan string, replaces newlines and carriage returns with spaces instead of empty strings. Fixes issue where successive route points get concatenated together, causing the Skyvector API to ignore those points.

- [ ] QC test passes locally (`make test`) (note: no changes in test results before vs after change)
- [ ] CI tests pass


trobarge26:
I am not sure why it is failing the integration tests. I've run the tests locally and get the same results before and after I implement the change.

PR #324, opened 2025-06-17
Fix flightplan string parser bug

trobarge26:
#### Fixes
Previously, when parsing flightplan strings containing newlines in the route, if the flight planning software does not add a space prior to the newline, the tokens adjoining the newline are concatenated together. This caused the Skyvector API to ignore the concatenated token, and thus both original tokens were inadvertently dropped from the route.

## Tests

- [x] QC test passes locally (`make test`) - no delta in results before and after change implemented
- [x] CI tests pass

## Reviewer

@zebengberg


zebengberg (approved):
Thanks!

trobarge26:
Updated Changelog.md to include this fix.

PR #330, opened 2025-08-08
Feature/landsat correction

ThymenW:
## Changes

- Add new Sentinel metadata downloading functions. Only valid for data after 2022 at the moment.
- Add algorithm to turn viewing angles into a pixel-level viewing angle dataset
- Compute the sensing_time offset due to the detector
- Add iterative scanning angle correction algorithm
- Add colocate_flights to calculate the exact sensing_time and aircraft location in the image. Including all corrections.
- Create Landsat | Sentinel function to use shared handler.get..... function in functions that should be shared.

## Tests

- [ ] CI tests pass

## Reviewer

@zebengberg

### DIFF
--- a/pycontrails/datalib/_leo_utils/correction.py
+++ b/pycontrails/datalib/_leo_utils/correction.py
@@ -112,6 +112,7 @@ def scan_angle_correction_iterative(
         dx_offset = offset * np.sin(vaa_rad)
         dy_offset = offset * np.cos(vaa_rad)
 
+        # Update the newly predicted x and y locations
         x_proj[update_indices] = x[update_indices] - dx_offset
         y_proj[update_indices] = y[update_indices] - dy_offset
 
@@ -310,23 +311,24 @@ def colocate_flights(flight: Flight, handler: Union[Sentinel, Landsat], utm_crs:
     # turn Flight object to dataframe
     flight_df = flight.dataframe.copy()
 
-    # keep only 30 seconds before and after sensing_time
+    # keep only 'search_window' minutes before and after sensing_time to speed up function
     flight_df["time"] = pd.to_datetime(flight_df["time"])
     flight_df = flight_df[flight_df["time"].between(initial_sensing_time - pd.Timedelta(minutes=search_window),
                                                      initial_sensing_time + pd.Timedelta(minutes=search_window))]
+    
     # add the x and y coordinates in the UTM coordinate system
     flight_df[['x', 'y']] = flight_df.apply(
         lambda row: pd.Series(geodetic_to_utm(row['longitude'], row['latitude'], utm_crs)), axis=1
     )
     
-    # add the x and y coordinates in the UTM coordinate system
+    # project the x and y location to the image level
     ds_viewing_angles = handler.get_viewing_angle_metadata()
     flight_df[['x_proj', 'y_proj']] = flight_df.apply(
         lambda row: pd.Series(scan_angle_correction_iterative(ds_viewing_angles, row['x'], row['y'], row["altitude"], iterations=3)), axis=1
     )
 
     # get the satellite ephemeris data prepared
-    satellite_ephemeris = handler.parse_ephemeris()
+    satellite_ephemeris = handler.get_ephemeris()
 
     # create the initial guess for the location 
     x_proj, y_proj = interpolate_columns(flight_df, initial_sensing_time, columns=["x_proj", "y_proj"])
@@ -343,7 +345,6 @@ def colocate_flights(flight: Flight, handler: Union[Sentinel, Landsat], utm_crs:
         if x_proj is None or y_proj is None or np.isnan(x_proj) or np.isnan(y_proj):
             raise ValueError("Interpolation failed after scan time correction iteration.")
         
-    # corrected_scan_time = pd.Timestamp(corrected_scan_time, tz='UTC')
     # Optional: correct sensing_time based on the detector it is in
     try:
         detector_id = handler.get_detector_id(x_proj, y_proj)
--- a/pycontrails/datalib/_leo_utils/correction.py
+++ b/pycontrails/datalib/_leo_utils/correction.py
@@ -1,6 +1,7 @@
 """Support for overlaying flight and contrail data on Landsat & Sentinel images."""
 
 import itertools
+from collections.abc import Sequence
 
 import numpy as np
 import numpy.typing as npt
@@ -8,11 +9,10 @@ import pandas as pd
 import pyproj
 import shapely
 import xarray as xr
-from typing import Sequence, Union
 
 from pycontrails import Flight
-from pycontrails.datalib.sentinel import Sentinel
 from pycontrails.datalib.landsat import Landsat
+from pycontrails.datalib.sentinel import Sentinel
 
 
 def _ephemeris_ecef_to_utm(ephemeris_df: pd.DataFrame, utm_crs: pyproj.CRS) -> pd.DataFrame:
@@ -47,10 +47,10 @@ def scan_angle_correction_iterative(
         The dataset containing the viewing azimuth angle (VAA)
         and viewing zenith angle (VZA) arrays.
     x : npt.NDArray[np.floating]
-        The x coordinates of the points to correct. Should be in the 
+        The x coordinates of the points to correct. Should be in the
         correct UTM coorinate system
     y : npt.NDArray[np.floating]
-        The y coordinates of the points to correct. Should be in the 
+        The y coordinates of the points to correct. Should be in the
         correct UTM coorinate system.
     z : npt.NDArray[np.floating]
         The z coordinates (altitude in meters) of the points to correct.
@@ -58,7 +58,7 @@ def scan_angle_correction_iterative(
     Returns
     -------
     tuple[npt.NDArray[np.floating], npt.NDArray[np.floating]]
-        The corrected x and y coordinates as numpy arrays in the UTM 
+        The corrected x and y coordinates as numpy arrays in the UTM
         coordinate system.
     """
     # Confirm that x is monotonically increasing and y is decreasing
@@ -69,7 +69,7 @@ def scan_angle_correction_iterative(
     if not np.all(np.diff(ds["y"]) < 0.0):
         msg = "ds['y'] must be monotonically decreasing"
         raise ValueError(msg)
-    
+
     x = np.atleast_1d(x).astype(np.float64)
     y = np.atleast_1d(y).astype(np.float64)
     z = np.atleast_1d(z).astype(np.float64)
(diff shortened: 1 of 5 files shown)


pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
+    x, y, h = transformer.transform(ecef_x, ecef_y, ecef_z)
+    return pd.DataFrame({"x": x, "y": y, "z": h, "t": ecef_t})
+
+
+def scan_angle_correction_iterative(
+    ds: xr.Dataset,
+    x: npt.NDArray[np.floating],
+    y: npt.NDArray[np.floating],
+    z: npt.NDArray[np.floating],
+    iterations: int = 5,
```
zebengberg: For purely stylistic reasons, you may want to rename `iterations` -> `n_iter` (we use `n_iter` in a few other places in pycontrails).

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
-    if match is None:
-        raise ValueError("No data found for EPHEMERIS_EPOCH_SECONDS in the ANG content.")
-    seconds = float(match.group(1))
-
-    t0 = (
-        pd.Timestamp(year=year, month=1, day=1)
-        + pd.Timedelta(days=day - 1)
-        + pd.Timedelta(seconds=seconds)
-    )
+    x = np.atleast_1d(x).astype(np.float64)
```
zebengberg: You could possibly avoid a copy with `.astype(np.float64, copy=False)`, but this is not a big deal!

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
+    x_proj[out_of_bounds] = np.nan
+    y_proj[out_of_bounds] = np.nan
 
-        data_list = [float(x.strip()) for x in data_str.split(",")]
-        if key == "EPHEMERIS_TIME":
-            data_list = [t0 + pd.Timedelta(seconds=t) for t in data_list]
-        arrays[key] = data_list
+    for _ in range(iterations):
+        # Use only valid points
+        valid = ~np.isnan(x_proj) & ~np.isnan(y_proj)
```
zebengberg: Might be cleaner to write this as `valid = np.isfinite(x_proj) & np.isfinite(y_proj)`.

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
+        x_proj[update_indices] = x[update_indices] - dx_offset
+        y_proj[update_indices] = y[update_indices] - dy_offset
+
+    return x_proj, y_proj
 
-    x, y, h = transformer.transform(ecef_x, ecef_y, ecef_z)
-    return pd.DataFrame({"x": x, "y": y, "z": h, "t": ecef_t})
+
+# Interpolation function using xarray
+def interpolate_angles(ds, xi, yi):
```
zebengberg: If only using this function inside `scan_angle_correction_iterative`, I'd suggest making it private: `_interpolate_angles`. You could also give it some type hints. I'd also suggest removing the first two lines of code: `xi = np.atleast_1d(xi).flatten()`. It's good to be defensive, but if we're only calling this from `scan_angle_correction_iterative` then we already know `xi` and `yi` are 1 dimensional numpy arrays.

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
+    """
+    Bilinear interpolation for VZA and VAA from ds at points (xi, yi).
+
+        -> method can be switched to linear for better accuracy
+    """
+    xi = np.atleast_1d(xi).flatten()
+    yi = np.atleast_1d(yi).flatten()
+
+    vza = ds["VZA"].interp(x=("x", xi), y=("y", yi), method="nearest").values
+    vaa = ds["VAA"].interp(x=("x", xi), y=("y", yi), method="nearest").values
```
zebengberg: For this function to be fully vectorized, I think you should change this to:

```python
vza = ds["VZA"].interp(x=xr.DataArray(xi, dims="points"), y=xr.DataArray(yi, dims="points"))
```

As written, you'll a value for each point in the product of `xi \times yi`, but we want an interpolated value for pair `(xi, yi)` (abusing notation). For scalar `xi` and `yi`, it doesn't matter. The xarray documentation does a descent job discussing this: https://docs.xarray.dev/en/stable/user-guide/interpolation.html#advanced-interpolation

You can make this even a little more performant if you interpolate both variables at once.

```python
interped = ds[["VZA", "VAA"]].interp(x=xr.DataArray(xx, dims="points"), y=xr.DataArray(yy, dims="points"))
return interped["VZA"].values, interped["VAA"].values
```

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
+        if not np.any(valid):
+            break
 
-    return pd.DataFrame(arrays)
+        # Interpolate angles only for valid points
+        vza, vaa = interpolate_angles(ds, x_proj[valid], y_proj[valid])
 
+        # Flatten and clean up any unexpected shapes
+        vza = np.atleast_1d(vza).astype(np.float64).flatten()
+        vaa = np.atleast_1d(vaa).astype(np.float64).flatten()
```
zebengberg, line 97: I'd remove unless it's necessary

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
 
-def _ephemeris_ecef_to_utm(ephemeris_df: pd.DataFrame, utm_crs: pyproj.CRS) -> pd.DataFrame:
-    # Define the source CRS: ECEF (Earth-Centered, Earth-Fixed) with WGS84 datum
-    source_crs = pyproj.CRS(proj="geocent", datum="WGS84")
+        # Skip any interpolated NaNs
+        interp_valid = ~(np.isnan(vza) | np.isnan(vaa))
```
zebengberg: So we'll use the original `x` value if the interpolated `x` value is nan? I think that makes sense, I just want to make sure I'm understanding.

If you like the previous suggestion, you can also use `np.isfinite` here.

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -10,94 +10,128 @@
...
 
-    # Create a transformer object to convert from source CRS to target CRS
-    # The default order for ECEF is (X, Y, Z) and for UTM is (Easting, Northing, Height)
-    transformer = pyproj.Transformer.from_crs(source_crs, utm_crs, always_xy=False)
+        # Full index into original arrays
+        valid_indices = np.where(valid)[0]
```
zebengberg: The right-hand side can be written as `np.flatnonzero(valid)`. If you don't like that, you should change `np.where` -> `np.nonzero` (see the NOTE here: https://numpy.org/doc/stable/reference/generated/numpy.where.html)

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -260,3 +295,153 @@ def estimate_scan_time(
...
+
+
+def _geodetic_to_utm(lon, lat, crs):
+    """Convert geographic coordinates (longitude, latitude) to UTM coordinates."""
+    transformer = pyproj.Transformer.from_crs("EPSG:4326", crs, always_xy=True)
+    x_utm, y_utm = transformer.transform(lon, lat)
+    return x_utm, y_utm
+
+
+def colocate_flights(
```
zebengberg, line 327: I keep misspelling colocate as collocate. Thank you!

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -260,3 +295,153 @@ def estimate_scan_time(
...
+    flight_df = flight_df[
+        flight_df["time"].between(
+            initial_sensing_time - pd.Timedelta(minutes=search_window),
+            initial_sensing_time + pd.Timedelta(minutes=search_window),
+        )
+    ]
+    # add the x and y coordinates in the UTM coordinate system
+    flight_df[["x", "y"]] = flight_df.apply(
+        lambda row: pd.Series(_geodetic_to_utm(row["longitude"], row["latitude"], utm_crs)), axis=1
+    )
```
zebengberg: I believe `pd.DataFrame.apply` is generally not vectorized (as in, there's a big python for-loop under the hood). But your function `_geodetic_to_utm` is itself vectorized. So you can rewrite this whole thing as:

```python
x, y = _geodetic_to_utm(flight_df["longitude"], flight_df["latitude"], utm_crs)
flight_df.loc[:, "x"] = x
flight_df.loc[:, "y"] = y
```

The same is true for the next `flight_df.apply(... scan_angle_correction_iterative)` as well.

pycontrails/datalib/landsat.py
```diff
@@ -264,6 +264,31 @@ def _get_meta(self) -> str:
...
+        pass
+
+    def get_crs(self):
+        """Return the CRS of the satellite image."""
+        pass
+
+    def get_sensing_time(self):
+        """Return the sensing_time of the satellite image."""
+        pass
+
```
zebengberg: Thanks! I can work on all these. For now, you may want to `raise NotImplementedError` instead of `pass`.

zebengberg (approved):
Thanks @ThymenW ! This looks great. Whenever you're ready, merge this into my branch and I can pick up from there.

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -260,3 +295,153 @@ def estimate_scan_time(
...
+                "No valid scan time could be estimated from the UTM coordinates during iteration."
+            )
+
+        corrected_sensing_time = corrected_sensing_time[0]
+
+        x_proj, y_proj = interpolate_columns(
+            flight_df, corrected_sensing_time, columns=["x_proj", "y_proj"]
+        )
+        if x_proj is None or y_proj is None or np.isnan(x_proj) or np.isnan(y_proj):
+            raise ValueError("Interpolation failed after scan time correction iteration.")
```
zebengberg: This is clever. I hadn't thought about doing this iteratively.

pycontrails/datalib/_leo_utils/correction.py
```diff
@@ -260,3 +295,153 @@ def estimate_scan_time(
...
+
+    timestamp = pd.to_datetime(timestamp)
+
+    if timestamp <= df[time_col].iloc[0]:
+        before, after = df.iloc[0], df.iloc[1]
+    elif timestamp >= df[time_col].iloc[-1]:
+        before, after = df.iloc[-2], df.iloc[-1]
+    else:
+        before = df[df[time_col] <= timestamp].tail(1).iloc[0]
+        after = df[df[time_col] >= timestamp].head(1).iloc[0]
```
zebengberg: Perhaps change to:

```python
before = df[df[time_col] <= timestamp].iloc[-1]
after = df[df[time_col] >= timestamp].iloc[0]
```

zebengberg (approved):

PR #341, opened 2025-09-25
Feature/himawari

zebengberg:
### Features

- Add new `Himawari` interface for downloading, processing, and visualizing [Himawari-8 and Himawari-9](https://www.jma.go.jp/jma/jma-eng/satellite/introduction.html) imagery. See the [Himawari notebook](https://py.contrails.org/notebooks/Himawari.html) for usage examples.

### Internals

- Add a new `pycontrails.datalib.geo_utils` module with tools for working with geostationary satellite imagery. This module currently includes functions previously defined in the `goes` module, but are also applicable to Himawari imagery.
- Rename `goes.extract_goes_visualization` -> `goes.extract_visualization` for consistency with the new `himawari` module. The old name is still available as an alias.
- Localize the `cartopy` import to the `extract_visualization` function within the `goes` module. This allows the user to download `GOES` imagery without having `cartopy` installed.
- Tweak some of the internal logic in `GOES.get` to slightly improve performance and readability, and to avoid loading unnecessary cached data into memory.

## Tests

- [x] CI tests pass

## Reviewer

@thabbott

### DIFF
--- a/pycontrails/datalib/himawari/himawari.py
+++ b/pycontrails/datalib/himawari/himawari.py
@@ -527,7 +527,15 @@ class Himawari:
 
 def _cartopy_crs(proj4_string: str) -> cartopy.crs.Geostationary:
     try:
-        import pyproj  # pyproj is a dependency of cartopy
+        import pyproj
+    except ModuleNotFoundError as exc:
+        dependencies.raise_module_not_found_error(
+            name="Himawari visualization",
+            package_name="pyproj",
+            module_not_found_error=exc,
+            pycontrails_optional_package="sat",
+        )
+    try:
         from cartopy import crs as ccrs
     except ModuleNotFoundError as exc:
         dependencies.raise_module_not_found_error(


thabbott:
I haven't done a full review yet--just posting comments on the new notebook.

Could we try to find a Japan or Target scene that includes some contrails? I see maybe one contrail in the upper left corner of the Target scene that you used, but it would be nice to be able to show a bunch of contrails and point out that they look basically the same as they do in GOES ash RGBs.

pycontrails/datalib/geo_utils.py
```diff
@@ -0,0 +1,261 @@
+"""Tooling and support for GEO satellites."""
```
thabbott, line 1: Do any of these functions need careful review, or were they just extracted from `goes.py` without changes?

zebengberg:
Thanks! Here's another nice example.

himawari-japan

pycontrails/datalib/himawari/himawari.py
```diff
@@ -0,0 +1,646 @@
...
+        return set(DEFAULT_BANDS)
+
+    if isinstance(bands, str):
+        bands = (bands,)
+
+    available = {f"B{i:02d}" for i in range(1, 17)}
+    bands = {b.upper() for b in bands}
+    if not bands.issubset(available):
+        raise ValueError(f"bands must be in {sorted(available)}")
+    return bands
```
thabbott, line 106: Should we just settle on using either "bands" or "channels" for all of our satellite datalibs? It seems silly to have a separate implementation of `goes._parse_channels` just to replace "channels" with "bands", and it could get annoying for users to have to keep track of which datalibs use which terminology.

(For what it's worth, "band" is more widely used, even for GOES--https://www.goes-r.gov/spacesegment/ABI-tech-summary.html)

pycontrails/datalib/himawari/himawari.py
```diff
@@ -0,0 +1,646 @@
...
+    return c0 + c1 * effective_bt + c2 * effective_bt**2
+
+
+def _radiance_to_reflectance(
+    radiance: npt.NDArray[np.float32],
+    calib_info: dict[str, Any],
+) -> npt.NDArray[np.float32]:
+    """Convert radiance to reflectance."""
+    coeff = calib_info["coeff_c_prime"]
+    return np.clip(radiance * coeff, 0.0, 1.0)
```
thabbott: Is this missing a correction offset? https://www.eorc.jaxa.jp/ptree/faq.html, Q3-2-7

Also, I don't think we should be clipping to (0, 1). If you're computing what GOES calls the "reflectance factor" (https://www.star.nesdis.noaa.gov/goesr/documents/ATBDs/Baseline/ATBD_GOES-R_ABI_CMI_KPP_v3.0_July2012.pdf) then the value's only guaranteed to be less than 1 if reflection is perfectly diffuse. It's often larger than 1 in imagery we get through the GOES datalib.

pycontrails/datalib/himawari/himawari.py
```diff
@@ -0,0 +1,646 @@
...
+        if len(da_dict) == 1 or "B03" not in da_dict:
+            return xr.concat(da_dict.values(), **kwargs).sortby("band_id")  # type: ignore[call-overload]
+        da03 = da_dict.pop("B03")
+        da1 = xr.concat(da_dict.values(), **kwargs)  # type: ignore[call-overload]
+        return geo_utils._coarsen_then_concat(da1, da03).sortby("band_id")
+
+
+def _cartopy_crs(proj4_string: str) -> cartopy.crs.Geostationary:
+    try:
+        import pyproj  # pyproj is a dependency of cartopy
```
thabbott: This doesn't necessarily mean that pyproj is guaranteed to be installed if cartopy is installed--I can create a environment with cartopy and no pyproj by running
```
pip install cartopy
pip uninstall pyproj
```
Not sure why anybody would do this... but maybe it's worth raising a separate error for a missing pyproj to avoid confusion?

pycontrails/datalib/himawari/himawari.py
```diff
@@ -0,0 +1,646 @@
...
+            "compat": "override",
+        }
+        if len(da_dict) == 1 or "B03" not in da_dict:
+            return xr.concat(da_dict.values(), **kwargs).sortby("band_id")  # type: ignore[call-overload]
+        da03 = da_dict.pop("B03")
+        da1 = xr.concat(da_dict.values(), **kwargs)  # type: ignore[call-overload]
+        return geo_utils._coarsen_then_concat(da1, da03).sortby("band_id")
+
+
+def _cartopy_crs(proj4_string: str) -> cartopy.crs.Geostationary:
```
thabbott, line 528: The rest of this file is code that's very nearly duplicated in `goes.py`. Is it worth extracting any more to `geo_utils.py`? (I'm not necessarily convinced one way or the other. If there are small differences that are hard to abstract away it's probably not worth it. There aren't that many geostationary satellites, so it's not like there's a risk of ending up with 10 copies of these functions.)

tests/unit/test_himawari.py
```diff
@@ -0,0 +1,87 @@
...
+        himawari_module._check_band_resolution(bands)
+        return
+
+    with pytest.raises(ValueError, match=r"Bands must have a common horizontal resolution"):
+        himawari_module._check_band_resolution(bands)
+
+
+@pytest.mark.skipif(OFFLINE, reason="offline")
+@pytest.mark.parametrize("t", ["2024-06-15T12:17:30", "2021-03-14T00:02:30"])
+def test_himawari_get_no_cache_default_bands(t: str) -> None:
```
thabbott, line 58: Not downloading any reflective bands means there's code for converting radiances to reflectances that isn't exercised in the himawari tests. (This isn't an issue for the GOES test suit since the conversion's already been done by the time we pull the data.) I didn't have any issues processing reflective bands when I was playing with the datalib, but I still think it's worth testing.

thabbott (approved):
I spent a bunch of time playing with this datalib and didn't run into any major issues! It is significantly slower (at least on my laptop) than the GOES datalib, I assume because more processing's happening on the user side?

I only found one technical issue; otherwise my comments are mostly about refactoring (more could maybe be extracted to `geo_utils.py`, but it's not obvious whether that's worth it) and testing.

pycontrails/datalib/geo_utils.py
zebengberg: Copied directly -- no need to review!

pycontrails/datalib/himawari/himawari.py
zebengberg: Yeah, I'd like to updates GOES to use `bands`. I think I'll do this and add some deprecation warnings.

zebengberg: 7a14be149b766a866f81df6da2ad901d5287e48a

tests/unit/test_himawari.py
zebengberg: I just added a quick test that downloads B02 and B03! It's really just a smoke test.

zebengberg: 96686f3d6fcc6702a4cd4f0b0439925acc60fd9e

pycontrails/datalib/himawari/himawari.py
zebengberg: Yeah, I definitely thought about that! There's a lot of repetition, but I agree with you that some redundancy is okay for now. (For example, we could have some ABC like `GEOSatellite` with an abstract `get` method ... it just seemed like overkill). I essentially moved what could be used directly in both `GOES` and `Himawari`, but I didn't want to go out of my way to add more. I think if we ever support an MTG interface that also uses some of these helper functions, that could be the right time to abstract.

pycontrails/datalib/himawari/himawari.py
zebengberg: 166b42c6df99ff2edf0667d17840635f9bbca8bd

pycontrails/datalib/himawari/himawari.py
zebengberg: I removed the clipping but kept the multiplicative-only correction (per our offline conversation and another Himawari document: https://www.data.jma.go.jp/mscweb/en/himawari89/space_segment/hsd_sample/HS_D_users_guide_en_v13.pdf). If we do notice weird values later on, we can revisit this.

Issue #249, opened 2024-10-02
GCPCacheStore fails when `cache_dir` is defined


mlshapiro:
**Description**

GCPCacheStore fails when `cache_dir` is defined. 

**Details**

- Version: 0.54
- Module: `cache`

**Steps to Reproduce**

```python
GCP_CACHE = GCPCacheStore(bucket=BUCKET, cache_dir="cache")
DATE = pd.Timestamp("2019-02-01")

time_bounds = (DATE, DATE + pd.Timedelta("36h"))
pressure_levels = (350, 300, 250, 225, 200, 175, 150)

era5pl = ERA5(
    time=time_bounds,
    variables=Cocip.met_variables,
    pressure_levels=pressure_levels,
    cachestore=GCP_CACHE
)
```

**Additional Notes**

We may not want to support the `GCPCacheStore` anymore given that we have Zarr files. Its convenient for demos, but

PR #346, opened 2025-10-20
Feature/gruan

zebengberg:
#### Features

- Add lightweight `GRUAN` datalib for accessing GRUAN radiosonde data via FTP. See the [GRUAN notebook](https://py.contrails.org/notebooks/GRUAN.html) for usage examples.

## Tests

- [x] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.59.0 (unreleased)
+## 0.59.0
 
 ### Features
 


pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+import tempfile
+
+import xarray as xr
+
+from pycontrails.core import cache
+
+#: GRUAN FTP server address
+FTP_SERVER = "ftp.ncdc.noaa.gov"
+
+#: All available GRUAN products and sites on the FTP server as of 2025-10
```
thabbott, line 20: How important is it for this to stay in sync with the output of `available_sites`?

If it's very important, then it might be a better idea just to use `available_sites`, even if there's a delay the first time it's called.

If it's convenient but not critical, can we add a test that will fail if this falls out of sync with the output of `available_sites`?

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+
+def extract_gruan_time(filename: str) -> tuple[datetime.datetime, int]:
+    """Extract launch time and revision number from a GRUAN filename."""
+    parts = filename.split("_")
+    if len(parts) != 6:
+        raise ValueError(f"Unexpected filename format: {filename}")
+    time_part = parts[4]
+    if len(time_part) != 15 or time_part[8] != "T":
+        raise ValueError(f"Unexpected time segment: {time_part}")
+    time = datetime.datetime.strptime(time_part, "%Y%m%dT%H%M%S")
```
thabbott:
```suggestion
    try:
       time = datetime.datetime.strptime(time_part, "%Y%m%dT%H%M%S")
    except:
       raise ValueError(f"Unexpected time segment: {time_part}")
```

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        "TEN",
+        "GVN",
+    ],
+    "RS92-PROFILE-BETA.2": ["BOU", "CAB", "LIN", "POT", "SOD", "TAT"],
+    "RS92-PROFILE-BETA.3": ["BOU", "CAB", "LIN", "POT", "SOD", "TAT"],
+}
+
+
+def extract_gruan_time(filename: str) -> tuple[datetime.datetime, int]:
+    """Extract launch time and revision number from a GRUAN filename."""
```
thabbott: Here and elsewhere in this file: either prefix with '_' if the function's meant to be private, or add a more detailed docstring if it's meant to be public?

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        - :meth:`list_files` -> list available NetCDF files for the given year
+        - :meth:`get` -> download and open a single NetCDF file as an :class:`xarray.Dataset`
+
+    Typical workflow:
+
+        1. Inspect :attr:`AVAILABLE` (fast) or call :func:`available_sites` (live)
+        2. Instantiate ``GRUAN(product, site)``
+        3. Call ``years()``
+        4. Call ``list_files(year)``
+        5. Call ``get(filename)`` for an ``xarray.Dataset``
```
thabbott, line 184: I have a prototype GMN datalib (currently in a private repo) that I've been using to get camera imagery from the GMN SFTP server. One thing that I don't like about the way that datalib works is that any call to its `get` method will result in a connection to the SFTP server, even if all the files it needs are already in the cache.

The reason it's hard to avoid this is that there's not a straightforward mapping from a GMN station ID and time to a filename on the GMN SFTP server, so you end up having to list and filter files on the server just to figure out what filenames you need. I think you're dealing with a similar problem here, and I think this solution's reasonable. Did you think about any alternatives, though?

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        self.product = product
+
+        if site not in known[product]:
+            known = available_sites()  # perhaps AVAILABLE_PRODUCTS_TO_SITES is outdated
+            if site not in known[product]:
+                raise ValueError(
+                    f"Unknown GRUAN site '{site}' for product '{product}'. "
+                    f"Known sites: {known[product]}"
+                )
+        self.site = site
```
thabbott, line 217: This seems fine as long as we're reasonably confident that `AVAILABLE_PRODUCTS_TO_SITES` will always be a subset of `available_sites()` (i.e., as long as we expect data to be added but not removed from the SFTP server, which I assume is how things normally work with GRUAN data).

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+                sites = ftp.nlst()
+
+                key = f"{name}.{int(version.split('-')[-1])}"
+                out[key] = sites
+
+    return out
+
+
+class GRUAN:
+    """Access `GRUAN <https://www.gruan.org/>`_ data over FTP.
```
thabbott:
```suggestion
    """Access `GRUAN <https://www.gruan.org/>`_ data over anonymous FTP.
```
(To make it clear that users don't need credentials?)

And are users supposed to register anywhere before they use GRUAN data? If so we should mention it here.

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        return f"{self.base_path_product}/{self.site}"
+
+    def years(self) -> list[int]:
+        """Get a list of available years for the selected product and site."""
+        ftp = self.connect()
+        ftp.cwd(self.base_path_site)
+        years = ftp.nlst()
+        return [int(year) for year in years]
+
+    def list_files(self, year: int) -> list[str]:
```
thabbott:
```suggestion
    def list_files(self, year: int | None = None) -> list[str]:
```
... the idea being for `self.list_files(2020)` to provide files for 2020 and `self.list_files()` just to provide a list of all the available files.

(This isn't super important, but might be a nice QOL addition.)

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        path = f"{self.base_path_site}/{t.year}/{filename}"
+
+        ftp = self.connect()
+        with tempfile.NamedTemporaryFile(delete=False) as tmp:
+            ftp.retrbinary(f"RETR {path}", tmp.write)
+
+        try:
+            return xr.load_dataset(tmp.name)
+        finally:
+            os.remove(tmp.name)
```
thabbott:
```suggestion
        with tempfile.NamedTemporaryFile(delete=True, delete_on_close=False) as tmp:
            ftp.retrbinary(f"RETR {path}", tmp.write)
            tmp.close()
            return xr.load_dataset(tmp.name)
```
I don't think the current code will clean up the temporary file if something goes wrong during `ftp.retrbinary`.

thabbott: Though I don't think the `delete_on_close` kwarg is available before python 3.12, so my solution might be worse...

pycontrails/datalib/gruan.py
```diff
@@ -0,0 +1,280 @@
...
+        lpath = self.cachestore.path(filename)
+        if self.cachestore.exists(lpath):
+            return xr.open_dataset(lpath)
+
+        t, _ = extract_gruan_time(filename)
+        path = f"{self.base_path_site}/{t.year}/{filename}"
+
+        ftp = self.connect()
+        with open(lpath, "wb") as f:
+            ftp.retrbinary(f"RETR {path}", f.write)
```
thabbott, line 341: Would it be better to write to a temporary file and add the file to the cachestore with `self.cache.put`? As is I'm not sure whether this behaves as expected when the cache is a writeable `GCPCacheStore`. (I don't think it will produce an error, since `lpath` should be a path from a local mirror of a GCS bucket, but the file won't be uploaded to GCS after being written locally.)

tests/unit/test_gruan.py
```diff
@@ -0,0 +1,120 @@
...
+
+    ftp = gruan.connect()
+    ftp.cwd(base_path)
+    products = ftp.nlst()
+    expected = {"RS41-EDT", "RS-11G-GDP", "RS92-GDP", "RS92-PROFILE-BETA"}
+    # exclude files with dots (readme.txt etc)
+    assert {p for p in products if "." not in p} == expected
+
+
+def test_years(gruan: GRUAN) -> None:
```
thabbott, line 73:
```suggestion
@pytest.mark.skipif(OFFLINE, reason="offline")
def test_years(gruan: GRUAN) -> None:
```

tests/unit/test_gruan.py
```diff
@@ -0,0 +1,120 @@
...
+    """Test available years retrieval."""
+    years = gruan.years()
+    # For this site/product, data is available 2005-2021
+    assert years == list(range(2005, 2022))
+
+
+def test_list_files_and_extract_time(gruan: GRUAN) -> None:
```
thabbott, line 81:
```suggestion
@pytest.mark.skipif(OFFLINE, reason="offline")
def test_list_files_and_extract_time(gruan: GRUAN) -> None:
```

docs/notebooks/GRUAN.ipynb
```diff
@@ -0,0 +1,2858 @@
+{
```
thabbott, line 1: <p>Could you use cartopy to add some coastlines or country outlines? That would give a better intuitive sense of how far the balloon travels. (I'd also find it more informative to color the points based on altitude rather than minutes since launch, since that makes it easier to figure out how far it moved horizontally in the FL300-400 range we mostly care about.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/contrailcirrus/pycontrails/pull/346/discussion/#comment-2449501899'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='10'/>

thabbott, line 1: <p>If you want to include this plot I think we should include some text emphasizing that the ISSRs at low altitude probably won't be PCRs.</p><p> </p><p>(Or, better yet, you could use GRUAN temperatures with some nominal engine efficiency to estimate what portion are PCRs and include that on this plot.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/contrailcirrus/pycontrails/pull/346/discussion/#comment-2449501901'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='15'/>

thabbott, line 1: <p>I'm not sure this is worth including... I think it's cool, but there are a lot of difficult questions about how to prescribe meteorology based on observations that people need to think through before they take experiments like this seriously.</p><p> </p><p>If you want to include a section inviting people to use this datalib for interesting work (I think you said something along those lines at standup), what about including a quick demo showing how to compare GRUAN soundings with forecast data in a <code>MetDataset</code>? (It's probably too much to do anything statistically meaningful, but a code snippet showing how to do the interpolation for a single profile would probably be a helpful first step for a lot of people.)  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/contrailcirrus/pycontrails/pull/346/discussion/#comment-2449501907'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='17'/>

pycontrails/datalib/gruan.py
zebengberg: e2b29e6036eaf60611b7876f1e8b2a7a65c8c229

pycontrails/datalib/gruan.py
zebengberg: In an initial version of this, I just recreated a new `ftplib.FTP` at each step (listing available files, downloading), but like you're describing, I found this slow and a bit wasteful. So I opted for the current implementation, which has drawbacks, but I think the benefits outweigh them.

I think it would be possible to download the entire GRUAN file tree and cache somewhere so that subsequent lookups would be fast, but it didn't feel like it was worth the effort.

pycontrails/datalib/gruan.py
zebengberg: No registration needed. Like you said, all data can be accessed anonymously.

pycontrails/datalib/gruan.py
zebengberg: > I don't think the current code will clean up the temporary file if something goes wrong during ftp.retrbinary.

Good point. I'll just wrap the whole thing in try ... finally

zebengberg: 13f8e4ec30d3726a4695400b18a742794638d9bc

pycontrails/datalib/gruan.py
zebengberg: The only reason I didn't test `available_sites` is because it's slow to build the recursive FTP tree. I'm going to spend some time to see if I can speed this up ....

zebengberg: To answer your question, it's totally fine if `AVAILABLE` goes out of sync with `available_sites`. I added a unit test to ensure they're equal, so if it fails, we can just update. If this gets annoying, we could just use `available_sites`, but I don't expect these to change often.

d495cb3e07f316a875158ea07f6637671106de3b

pycontrails/datalib/gruan.py
zebengberg: Yes, I'm sure right ... but I also think we may remove `GCPCacheStore` at some point, so I don't really want to consider it here. https://github.com/contrailcirrus/pycontrails/issues/249

docs/notebooks/GRUAN.ipynb
zebengberg: <p>Okay! I opted to jettison this whole section for simplicity. If we need to reference it, it's in the git history.</p><p> </p><p>364275467e7522d1ae35f4047d1f992c308aed69</p>

zebengberg: <p>I went for it and computed PCR instead.</p><p> </p><p>364275467e7522d1ae35f4047d1f992c308aed69</p>

zebengberg: <p>364275467e7522d1ae35f4047d1f992c308aed69</p>

pycontrails/datalib/gruan.py
zebengberg: 66c2b53ee4e291a404398b2f8427eb0831f62f7d

PR #352, opened 2025-10-31
IAGOS datalib

thabbott:
#### Features

- Add experimental IAGOS datalib. This datalib is currently untested and requires raw IAGOS data files to be available locally.
- Add optional time filtering to GRUAN datalib.

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/pycontrails/datalib/iagos.py
+++ b/pycontrails/datalib/iagos.py
@@ -663,7 +663,7 @@ class IAGOS:
 
             if response.status_code == 401:  # could be using expired token
                 self._refresh_token()
-                session.headers.update({f"authorizationbearer {self.token}"})
+                session.headers.update({"authorization": f"bearer {self.token}"})
                 response = session.get(url, params=params)
 
             try:


pycontrails/datalib/gruan.py
```diff
@@ -266,17 +277,32 @@ def list_files(self, year: int | None = None) -> list[str]:
...
+                date = extract_gruan_time(file)[0]
+                return date >= self.timespan[0] and date <= self.timespan[1]
+
+            return sorted(
+                file
+                for y in years
+                if y >= self.timespan[0].year and y <= self.timespan[1].year
+                for file in self.list_files(y)
+                if _in_range(file)
+            )
```
zebengberg, line 307: Mostly cosmetic, but something like this would help convince me that `self.timespan` is what we think it should be (I had to read through the source code of `metsource.parse_timesteps` to understand that `self.timespan` is always a length-2 list. The suggestion below would fail if not.)

```suggestion

            t0, t1 = self.timespan

            def _in_range(file: str) -> bool:
                date = extract_gruan_time(file)[0]
                return date >= t0 and date <= t1

            return sorted(
                file
                for y in years
                if y >= t0.year and y <= t1.year
                for file in self.list_files(y)
                if _in_range(file)
            )
```

pycontrails/datalib/iagos.py
```diff
@@ -0,0 +1,429 @@
...
+        flight_ids.append(flight_id)
+        validated.append(str(path))
+
+    duplicates = sorted([f for f in set(flight_ids) if flight_ids.count(f) > 1])
+    if duplicates:
+        msg = (
+            f"IAGOS flight ids {duplicates} appear more than once in `paths`. "
+            f"Deduplicate files included in `paths` before using."
+        )
+        raise ValueError(msg)
```
zebengberg: Ignorable:

Computing `duplicates` has quadratic runtime. You might just consider the simpler:

```
if len(set(flight_ids)) != len(flight_ids):
```

then compute `duplicates` only in the case when you raise.

pycontrails/datalib/iagos.py
```diff
@@ -0,0 +1,429 @@
...
+
+    return validated
+
+
+class IAGOS:
+    """EXPERIMENTAL: class for processing L2 IAGOS data.
```
zebengberg: It would be helpful to provide a link to IAGOS data in the docstring here

pycontrails/datalib/iagos.py
```diff
@@ -0,0 +1,429 @@
...
+            the expected format.
+        """
+        if self.paths:
+            return sorted([os.path.basename(p) for p in self.paths])
+
+        msg = (
+            "File retrieval from the IAGOS database has not been implemented. "
+            "To use this datalib, provide paths to local IAGOS files using the `paths` parameter."
+        )
+        raise NotImplementedError(msg)
```
zebengberg: Are a lot of these methods written with the hope of supporting direct download from IAGOS in the future?

pycontrails/datalib/iagos.py
```diff
@@ -0,0 +1,429 @@
...
+        }
+        for variable in self.variables:
+            name = variable.standard_name
+            key = present[name]
+            flight[name] = ds[key]
+            flight[f"{name}_validity_flag"] = ds[f"{key}_validity_flag"]
+            if variable not in IAGOS_VARIABLES:
+                continue
+            flight[f"{name}_standard_error"] = ds[f"{key}_error"]
+            flight[f"{name}_process_flag"] = ds[f"{key}_process_flag"]
```
zebengberg: So what's going on here? Some variables have some additional information (standard error and process flag)?

thabbott: Yes. It seems like different metadata is available for variables that are measured by the aircraft directly (everything in `AIRCRAFT_VARIABLES`) and those measured by the onboard IAGOS instrument (`IAGOS_VARIABLES`, which right now just has water vapor). The error and processing flag are only available for the latter.

thabbott: b7a448b7a145d815b3ef8edb914cb000e2a6d76

pycontrails/datalib/iagos.py
thabbott: Yes. I got some information about authentication for the IAGOS API, so this should be straightforward to support!

pycontrails/datalib/iagos.py
thabbott: 7096530ce8056f68312aa62734b27fec93cd5faf

pycontrails/datalib/iagos.py
thabbott: 9d508b03d9daff6a1e12bdafdc43cc316f537d60

PR #356, opened 2025-11-10
Feature/meem v2

roger-teoh:
Closes #XX

## Changes

- Update ICAO EDB from v29b to v31.
- Add MEEM2 and SCOPE11 as alternative models to simulate nvPM emissions
- - Include more fields in the ICAO EDB static files. For example, MEEM2 utilises a fifth data point (maximum nvPM EI) to construct the nvPM emissions profile. 
- Major re-structuring and clean-up of the emissions module to improve intuitiveness and standardise terminology

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/pycontrails/models/emissions/emissions.py
+++ b/pycontrails/models/emissions/emissions.py
@@ -76,8 +76,8 @@ class Emissions(Model):
 
     See Also
     --------
-    :mod:`pycontrails.models.emissions.black_carbon`
-    :mod:`pycontrails.models.emissions.ffm2`
+    :mod:`pycontrails.models.emissions.nvpm`
+    :mod:`pycontrails.models.emissions.gaseous`
     """
 
     name = "emissions"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 0.59.1
 
 ### Features
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -220,8 +220,8 @@ Emissions
     :toctree: api/
 
     models.emissions.Emissions
-    models.emissions.black_carbon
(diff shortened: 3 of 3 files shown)


pycontrails/models/emissions/emissions.py
```diff
@@ -203,7 +204,9 @@ def eval(self, source: GeoVectorDataset | None = None, **params: Any) -> GeoVect
                 )
 
         self._gaseous_emission_indices(engine_uid)
-        self._nvpm_emission_indices(engine_uid)
+        # TODO: Zeb, how do you let the user choose between GAIA and MEEM?
```
roger-teoh: @zebengberg How do you let the user choose between the GAIA and MEEM approach for simulating nvPM emissions?

roger-teoh: We are also awaiting unit tests from Boeing to test the MEEM2 model. A quick sanity check suggests that both MEEM2 and SCOPE11 models are working fine.

roger-teoh: It also seems like these functions `_check_edb_gaseous_availability` and `_check_edb_nvpm_availability` in emissions.py are no longer used anywhere anymore. Should we get rid of it?

zebengberg: > @zebengberg How do you let the user choose between the GAIA and MEEM approach for simulating nvPM emissions?

This could be a parameter that is passed into the `Emissions` model. Then `Cocip` could take an `emissions` param as well.

zebengberg: > It also seems like these functions `_check_edb_gaseous_availability` and `_check_edb_nvpm_availability` in emissions.py are no longer used anywhere anymore. Should we get rid of it?

Yes, please remove.

PR #358, opened 2025-11-18
ICON datalib

thabbott:
#### Features

- Add a datalib to support ICON forecast access via the [DWD Open Data Server](https://opendata.dwd.de/). See the [ICON notebook](https://py.contrails.org/notebooks/ICON.html) for usage examples.

#### Internals

- Add `"DWD"` as a recognized meteorology provider and `"ICON"`, `"ICON-EU"`, and `"ICON-D2"` as recognized meteorology datasets in `met.py`.
- Add geometric (height-coordinate) vertical velocity as a new generic `MetVariable`.
- Modify `metsource.parse_timesteps` to accept a new optional `shift` argument. This argument can be used to make `metsource.parse_timesteps` return times that are not even multiple of `freq` (e.g., six-hourly timesteps starting at 3Z). The behavior of `metsource.parse_timesteps` is unchanged if this argument is omitted.
- Extract some utilities for ECMWF model-to-pressure-level conversion to a more flexible (less ECMWF-specific) `met_utils.ml_to_pl` function and a new `utils.arrays` module. The updated utilities include some minor behavioral changes to improve handling of nan values.

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg

### DIFF
--- a/pycontrails/datalib/dwd/icon.py
+++ b/pycontrails/datalib/dwd/icon.py
@@ -441,9 +441,6 @@ class ICON(metsource.MetDataSource):
         If True, cache downloaded GRIB files rather than storing them in a temporary file.
         By default, False.
 
-    download_threads : int, optional
-        Limit the number of threads used to download data.
-
 
     See Also
     --------
@@ -459,7 +456,6 @@ class ICON(metsource.MetDataSource):
         "cache_download",
         "cachestore",
         "domain",
-        "download_threads",
         "forecast_time",
         "model_levels",
         "show_progress",
@@ -480,9 +476,6 @@ class ICON(metsource.MetDataSource):
     #: Whether to save raw data files to :attr:`cachestore` for reuse
     cache_download: bool
 
-    #: Number of threads used to download data
-    download_threads: int | None
-
     def __init__(
         self,
         time: metsource.TimeInput,
@@ -496,7 +489,6 @@ class ICON(metsource.MetDataSource):
         show_progress: bool = False,
         cachestore: cache.CacheStore = __marker,  # type: ignore[assignment]
         cache_download: bool = False,
-        download_threads: int | None = None,
     ) -> None:
         # Parse and set instance attributes
 
@@ -583,7 +575,6 @@ class ICON(metsource.MetDataSource):
         self.show_progress = show_progress
         self.cachestore = cache.DiskCacheStore() if cachestore is self.__marker else cachestore
         self.cache_download = cache_download
-        self.download_threads = download_threads
         self._global_kdtree: KDTree | None = None
 
     def __repr__(self) -> str:


pycontrails/datalib/_met_utils/metsource.py
```diff
@@ -68,6 +76,21 @@ def parse_timesteps(time: TimeInput | None, freq: str | None = "1h") -> list[dat
...
+    [datetime.datetime(2000, 1, 1, 0, 0)]
+
+    >>> parse_timesteps(("2000-01-01", "2000-01-02"), freq=None)
+    [datetime.datetime(2000, 1, 1, 0, 0), datetime.datetime(2000, 1, 2, 0, 0)]
+
+    >>> parse_timesteps(("2000-01-01 00:00", "2000-01-01 03:00"), freq="3h")
+    [datetime.datetime(2000, 1, 1, 0, 0), datetime.datetime(2000, 1, 1, 3, 0)]
+
+    >>> parse_timesteps(("2000-01-01 01:00", "2000-01-01 04:00"), freq="3h", shift="1h")
+    [datetime.datetime(2000, 1, 1, 1, 0), datetime.datetime(2000, 1, 1, 4, 0)]
```
zebengberg: All of these examples look like `f(t) = [t]` or `f((t0, t1)) = [t0, t1]`. It could be good to slightly tweak for some variety. For example, you could change the last one to:

```
>>> parse_timesteps(("2000-01-01 00:00", "2000-01-01 06:00"), freq="3h", shift="1h")
```

pycontrails/datalib/_met_utils/metsource.py
```diff
@@ -94,28 +117,46 @@ def parse_timesteps(time: TimeInput | None, freq: str | None = "1h") -> list[dat
 
     if freq is None:
         daterange = pd.DatetimeIndex([t0, t1])
-    else:
-        # get date range that encompasses all whole hours
+
+    elif shift is None:
```
zebengberg, line 123: So if `freq` is None, `shift` is ignored? If so, that could be added to the docstring.

pycontrails/datalib/dwd/icon.py
```diff
@@ -0,0 +1,1040 @@
...
+
+_met_var_to_ods_mapping = {
+    met_var.AirTemperature: "t",
+    met_var.SpecificHumidity: "qv",
+    met_var.NorthwardWind: "u",
+    met_var.EastwardWind: "v",
+    met_var.GeometricVerticalVelocity: "w",
+    met_var.MassFractionOfCloudIceInAir: "qi",
+    met_var.TOAOutgoingLongwaveFlux: "athb_t",
+    met_var.TOANetDownwardShortwaveFlux: "asob_t",
```
zebengberg, line 83: Wow, so these are different from the short names in some cases? That's confusing.

pycontrails/datalib/dwd/icon.py
```diff
@@ -0,0 +1,1040 @@
...
+    The DWD provides ICON forecasts on
+    `three domains <https://www.dwd.de/EN/ourservices/nwp_forecast_data/nwp_forecast_data.html>`_:
+    a global domain (~13 km resolution), a higher-resolution Europe domain (~7 km), and a
+    high-resolution Germany domain (~2 km). Global forecasts are initialized every 6 hours and
+    Europe and Germany forecasts are initialized every 3 hours. The Open Data Server *does not
+    provide a long-term forecast archive*. Data for each forecast cycle is typically available
+    for about 24 hours.
+
+    Global ICON forecasts are provided on an icosahedral grid and must be interpolated to a
+    latitude-longitude grid. *This datalib currently supports nearest-neighbor interpolation only.*
```
zebengberg: > *This datalib currently supports nearest-neighbor interpolation only.*

The rational you described for this (not wanting to smooth out high humidity regions via linear interpolation) really resonated with me ... do you want to include that here in the docstring?

pycontrails/datalib/dwd/icon.py
```diff
@@ -0,0 +1,1040 @@
...
+          - N/A
+        * - Extended forecast timestep
+          - 3h
+          - 3h
+          - 3h
+          - 6h
+          - N/A
+
+    This datalib currently supports only those variables required to run :class:`Cocip`
+    and :class:`CocipGrid`. Please
```
zebengberg, line 380: We could also run `DryAdvection`, `SAC`, and `ISSR` right? No need to change, just want to make sure.

pycontrails/datalib/dwd/icon.py
```diff
@@ -0,0 +1,1040 @@
...
+    --------
+    :func:`pycontrails.datalib.dwd.ods.list_forecasts`: list available forecast cycles
+    :func:`pycontrails.datalib.dwd.ods.list_timesteps`: list available forecast timesteps
+
+    """
+
+    __marker = object()
+
+    __slots__ = (
+        "__dict__",
```
zebengberg: What's the purpose of including `__dict__` in `__slots__`? Is this necessary?

pycontrails/datalib/dwd/icon.py
```diff
@@ -0,0 +1,1040 @@
...
+        self,
+        time: metsource.TimeInput,
+        variables: metsource.VariableInput,
+        pressure_levels: metsource.PressureLevelInput | None = None,
+        domain: str = "global",
+        timestep_freq: str | timedelta | None = None,
+        grid: float | None = None,
+        forecast_time: DatetimeLike | None = None,
+        model_levels: list[int] | None = None,
+        progress: bool = False,
```
zebengberg: We use `show_progress` elsewhere in pycontrails. Either name is fine, but it's probably better to be consistent, so I'd suggest changing this or changing the others.

pycontrails/utils/array.py
```diff
@@ -0,0 +1,95 @@
...
+    # The value chosen to replace nans doesn't matter,
+    # since indices are set to 0 for rows that contain
+    # any nans before returning.
+    #
+    # Ideally we'd like to find a non-nan value using
+    # a function that runs (best case) in O(1) rather
+    # than O(n) time... but I think this is probably
+    # faster than manually looping over a.
+    na_mask = np.isnan(a).any(axis=1, keepdims=True)
+    a = np.nan_to_num(a, nan=np.nanmax(a))
```
zebengberg: What about guarding some of the nan checking with something like:

```
contain_nans = na_mask.any()
if contains_nans:
    a = np.nan_to_num(a, nan=np.nanmax(a))

...

if contain_nans:
    return np.where(na_mask, 0, idx_scaled - n * steps.astype(np.int64))
return idx_scaled - n * steps.astype(np.int64)
```

It's probably not a big deal, but I really did try to optimize this for performance with ECMWF model level data and extra nan masking might slow it down a touch (or not, in which case ignore).

pyproject.toml
```diff
@@ -67,6 +67,7 @@ dev = [
     "pytest-cov>=2.11",
     "requests>=2.25",
     "ruff>=0.9.0",
+    "types-requests>=2.25",
```
zebengberg: What's this one for?

zebengberg (approved):
Thanks Tristan! This all looks great. I want to play a bit / test the ICON datalib locally before you merge ... I'll do this tomorrow.

I created some minor conflicts in the changelog you'll need to resolve.

pycontrails/datalib/dwd/ods.py
```diff
@@ -0,0 +1,314 @@
...
+
+async def _get_async(rpath: str, lpath: str) -> None:
+    """Async helper function for get."""
+    try:
+        async with (
+            aiohttp.ClientSession(raise_for_status=True) as session,
+            session.get(f"https://{rpath}") as response,
+        ):
+            content = await response.read()
+    except aiohttp.web.HTTPException as e:
```
zebengberg: I think this should be `aiohttp.ClientError` here (and one other place). It looks like `aiohttp.web.HTTPException` is something that a web server would raise instead of something that the client would catch.

pycontrails/datalib/dwd/ods.py
```diff
@@ -0,0 +1,314 @@
...
+    return coroutines.run(_get_async(rpath, lpath))
+
+
+async def _get_async(rpath: str, lpath: str) -> None:
+    """Async helper function for get."""
+    try:
+        async with (
+            aiohttp.ClientSession(raise_for_status=True) as session,
+            session.get(f"https://{rpath}") as response,
+        ):
```
zebengberg: Something else to consider here is creating the `ClientSession` upstream then passing it in here. This would improve performance by a hair (probably not noticeable), but I think it's considered better practice for async requests to create a single connection so that the SSL handshake takes place only once. So something like this:

```python
async def _ls_async(url: str, session: aiohttp.ClientSession | None) -> AsyncGenerator[str, None]:
    """Async helper function for _ls."""
    if session is None:
        local_session = True
        session = aiohttp.ClientSession(raise_for_status=True)
    else:
        local_session = False

    try:
        async with session.get(f"https://{url}") as response:
            text = await response.text()
    except aiohttp.ClientError as e:
        msg = f"Error listing contents of {url}."
        raise RuntimeError(msg) from e
    finally:
        if local_session:
            await session.close()
    ...
```

Then your `_list_forecasts_async` would look more like this:

```python
async def _list_forecasts_async(domain: str) -> list[datetime]:
    """Async helper function for _list_forecasts."""
    start_times = []

    async with aiohttp.ClientSession(raise_for_status=True) as session:
        cycles = [c async for c in _ls_async(_root(domain), session=session)]
        tasks = [anext(_ls_async(f"{cycle}/athb_t", session=session)) for cycle in cycles]

    ...
```

You could make the same change in `_get_async` as well.

pycontrails/datalib/dwd/ods.py
thabbott: 42779185ef5919180882b38398fd0ce7c13869bb

pyproject.toml
thabbott: It provides type stubs for the requests module. (Mypy knows about them and complains if you use requests without having type stubs available.) We don't need it now that we're using aiohttp instead of requests, though.

pycontrails/datalib/dwd/ods.py
thabbott: https://github.com/contrailcirrus/pycontrails/pull/358/commits/55ca7d73234e686c0208768e106c592eecf86de1

pycontrails/datalib/dwd/icon.py
thabbott: Yes!

pycontrails/datalib/dwd/icon.py
thabbott: It's used by `functools.cached_property`, which I'm using for the k-d tree for horizontal remapping. You could come up with some workaround that doesn't use `__dict__`, e.g.
```python
def set_kdtree(self) -> None:
   if self._global_kdtree:
      return
   # set up k-d tree
```
I don't have a strong opinion about one approach vs the other. I think what I have keeps most of the performance benefits from using `__slots__`, though you have to be careful not to accidentally add a bunch of attributes to `__dict___`. (You won't get an error if you try to assign to an attribute that's not in `__slots__`; it will just go into `__dict__` instead.)

zebengberg: Okay, that all makes sense! I have a slight preference for something like the `self._global_kdtree` approach you mentioned, or even removing `__slots__` entirely, but either way is good.

pycontrails/datalib/_met_utils/metsource.py
thabbott: f5cf08d64fb0b5902bf286f3556ebdd305ef404e

pycontrails/datalib/_met_utils/metsource.py
thabbott: f5cf08d64fb0b5902bf286f3556ebdd305ef404e

pycontrails/datalib/dwd/icon.py
thabbott: Yes... and in some cases the variable name inside the grib file is different from the variable name in the grib filename. But I think part of the job of a datalib is to handle this sort of translation so that the user eventually ends up with variables labeled with the expected shortnames. We do this with global dictionaries in a few different datalibs, and I think that's reasonably clear and easy to maintain/extend.

pycontrails/datalib/dwd/icon.py
thabbott: 333552157a8e0febd1c0a6c604bb4b7060943cb7

pycontrails/datalib/dwd/icon.py
thabbott: > I have a slight preference for something like the self._global_kdtree approach you mentioned

Ok, we can go with that, then.

thabbott: a0f3fa1aab652c09291409129a241b195ae13228

pycontrails/datalib/dwd/icon.py
thabbott: 83bd27c8c3507fbec46fafa72e057a2d3eb7fc8f

pycontrails/utils/array.py
thabbott: 05f1c836558e015a2e9c19605f1f042629cb362f

PR #370, opened 2026-01-30
WIP: Forecast API datalib

max-vogler:
#### Features

 * Add `ForecastApi` datalib to load precomputed forecast grids from contrail APIs like [Google](https://developers.google.com/contrails).

## Tests

- [X] Test passes locally (`make test`)
- [x] More tests to be added

## Reviewer

@nickmasson 

Hey, here is a work-in-progress of adding a `ForecastApi` datalib. Happy to receive some early, high-level feedback before finalizing & polishing this PR. I've chosen to implement `MetVariables` for the API data variables. The datalib design is inspired by `Model` (but not extending this class). For now, only VectorDataset of `time` is supported as input, but I'm planning to also support GeoVectorDataset in the future.

### DIFF
--- a/pycontrails/datalib/google_forecast.py
+++ b/pycontrails/datalib/google_forecast.py
@@ -32,16 +32,17 @@ Severity = MetVariable(
     standard_name="contrails",
     long_name="Contrail Severity Index",
     description="The severity (0-4) of forecasted contrail warming.",
-    units="1",
+    units="1",  # dimensionless.
 )
 
 
-EffectiveEnergyForcing = MetVariable(
+ExpectedEffectiveEnergyForcing = MetVariable(
     short_name="eeef_per_m",
     standard_name="expected_effective_energy_forcing",
     long_name="Expected Effective Energy Forcing",
     description=(
-        "The expected effective energy forcing of contrail warming (see https://developers.google.com/contrails/v1/forecast-description)."
+        "The expected effective energy forcing of contrail warming "
+        "(see https://developers.google.com/contrails/v1/forecast-description)."
     ),
     units="J m**-1",
 )
@@ -137,7 +138,7 @@ class GoogleForecast(metsource.MetDataSource):
     @property
     def supported_variables(self) -> list[MetVariable]:
         """Get supported variables."""
-        return [Severity, EffectiveEnergyForcing]
+        return [Severity, ExpectedEffectiveEnergyForcing]
 
     def create_cachepath(self, t: datetime.datetime) -> str:
         """Return cachepath to local data file based on datetime."""


pycontrails/datalib/forecast.py
```diff
@@ -0,0 +1,175 @@
...
+    description="The severity (0-4) of forecasted contrail warming.",
+)
+
+
+EnergyForcing = MetVariable(
+    short_name="ef_per_m",
+    standard_name="expected_effective_energy_forcing",
+    long_name="Expected Effective Energy Forcing",
+    description="The effective energy forcing of contrail warming.",
+)
```
zebengberg: I'd suggest renaming this to `EffectiveEnergyForcing` and changing the short name to something like `eef_per_m`. pycontrails already uses `ef_per_m` to be for (non-effective) energy forcing per distance. Please also include units.

pycontrails/datalib/forecast.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+    def __init__(
+        self,
+        params: GoogleForecastParams,
+        met_variables: tuple[MetVariable, ...] = (Severity,),
+    ) -> None:
+        self._params = params
+        self.met_variables = met_variables
+        super().__init__()
```
zebengberg: Tristan emailed you a bunch of suggestion for structuring this for similarity with some of the other pycontrails datalibs. It will break some of the patterns you've laid out here, but I would recommend you try to stay consistent. See the `ERA5` datalib for an example.

https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/ecmwf/era5.py

In particular, the use will need to define the target times here in `__init__`

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,260 @@
...
+    long_name="Contrail Severity Index",
+    description="The severity (0-4) of forecasted contrail warming.",
+)
+
+
+EffectiveEnergyForcing = MetVariable(
+    short_name="eef_per_m",
+    standard_name="expected_effective_energy_forcing",
+    long_name="Expected Effective Energy Forcing",
+    description="The effective energy forcing of contrail warming. It has the ERF/RF ratio applied.",
```
zebengberg: What's the unit here? J/m? Could be nice idea to include.

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,260 @@
...
+)
+
+
+class GoogleForecast(metsource.MetDataSource):
+    """Google Forecast datalib to download precomputed contrail forecasts from API sources.
+
+    This class provides an interface to the Google Contrails Forecast API.
```
zebengberg:
```suggestion
    This class provides an interface to the `Google Contrails Forecast API <https://developers.google.com/contrails>`_.
```

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,260 @@
...
+    def __init__(
+        self,
+        time: metsource.TimeInput | None,
+        variables: metsource.VariableInput = (Severity,),
+        key: str | google.auth.credentials.Credentials | None = None,
+        url: str = "https://contrails.googleapis.com/v2/grids",
+        cachestore: cache.CacheStore | None = None,
+    ) -> None:
+        self.url = url
+        self.cachestore = cachestore or cache.DiskCacheStore()
```
zebengberg: I think we want to set the cachestore directory. You could essentially just use a pattern like this one:

https://github.com/contrailcirrus/pycontrails/blob/e8a9268a7d46f04a551b8425029a7c419ed69690/pycontrails/datalib/gruan.py#L219-L223

and change the cache directory to

```python
cache_dir = f"{cache_root}/google-forecast"
```

max-vogler: Thanks! I have an open question about the cachestore behavior: In my initial prototype code I required the cachestore for simplicity. The datalib always downloads to the DiskCacheStore, then reads the NetCDF from there.

By default this could be unexpected behavior for users: On many OS, the cache folder is pretty much permanent (unlike temp). Thus, without explicit configuration from the user, pycontrails would easily use gigabytes of disk space.

I'll stick to project style (whatever you recommend), but from my intuition would propose disable the cache store by default and only keep NetCDF in memory. WDYT?

zebengberg: Yeah, we've thought about this in the past. Some of our datalibs allow the user to download data without caching to disk by setting `cachestore=None`. See the `GOES` datalib for an example of how we do this: https://github.com/contrailcirrus/pycontrails/blob/e8a9268a7d46f04a551b8425029a7c419ed69690/pycontrails/datalib/goes.py#L553-L555

In particular, see the `_get_without_cache` method there. It'd be great if you want to implement something like that here as well. Feel free to copy some of that logic or even abstract things into some common helper utilities that both `GoogleForecast` and `GOES` use. (I'm not sure if it's worth it, I haven't dug in.)

I also want to point out that Google is recomputing their contrail forecast every 6 hours (I think?) after ingesting the next ECMWF HRES forecast. I'm guessing that the Google API always serves the latest contrail forecast available. For historic forecasts, caching is fine because those forecasts are never stale. But for real time forecasts, perhaps it's better to avoid caching all together (or include some cache invalidation). However, selective caching would definitely complicate the pycontrails logic. You should decide if / how you want to include any of this into your datalib.

max-vogler: That's helpful! I've updated my code to not use any cache by default. This prevents staleness surprises for future forecasts. I've iterated the caching behavior for an edge case with multiple data variables: If at least one is missing for a timestamp, all are (re-)downloaded for this timestamp to ensure per-hour-consistency.

pycontrails/datalib/google_forecast.py
max-vogler: Added units. Yes, it is J/m.

max-vogler:
Hey @zebengberg and @nickmasson, I've finalized my code, please take another look! :)

max-vogler:
Zeb, google.auth.default() still is helpful with other forms of authentication, e.g. service accounts. We'll also try to rollout OAuth support in the next ~week. Would you still want this removed, or do we keep it?

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+Severity = MetVariable(
+    short_name="contrails",
+    standard_name="contrails",
+    long_name="Contrail Severity Index",
+    description="The severity (0-4) of forecasted contrail warming.",
+    units="dimensionless",
```
thabbott:
```suggestion
    units="1",
```

(The `units` fields is supposed to follow CF conventions, though I'm not sure how consistent we are about this elsewhere in pycontrails.)

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+)
+
+
+EffectiveEnergyForcing = MetVariable(
```
thabbott: I think we should call this `ExpectedEffectiveEnergyForcing`. It's a mouthful, but my understanding is that this is reduced relative to the instantaneous RF that CoCiP provides by the ERF/RF ratio (the "effective" part) and by the ML-based probability of contrail formation (the "expected") part. Not including the "Expected" part could be a pain in the future if we want to add an interface to some other data (e.g., from GCM simulations) that provides time- and space-integrated ERF.

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+    short_name="eef_per_m",
+    standard_name="expected_effective_energy_forcing",
+    long_name="Expected Effective Energy Forcing",
+    description=(
+        "The effective energy forcing of contrail warming. It has the ERF/RF ratio applied."
```
thabbott:
```suggestion
        "The expected effective energy forcing of contrail warming (see https://developers.google.com/contrails/v1/forecast-description)."
```

You already have a nice description of what your forecast returns--could we just point users there?

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+    short_name="contrails",
+    standard_name="contrails",
+    long_name="Contrail Severity Index",
+    description="The severity (0-4) of forecasted contrail warming.",
+    units="dimensionless",
+)
+
+
+EffectiveEnergyForcing = MetVariable(
+    short_name="eef_per_m",
```
thabbott:
```suggestion
    short_name="eeef_per_m",
```

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+    ),
+    units="J m**-1",
+)
+
+
+class GoogleForecast(metsource.MetDataSource):
+    """Google Forecast datalib to download precomputed contrail forecasts from API sources.
+
+    This class provides an interface to the `Google Contrails Forecast API <https://developers.google.com/contrails>`_.
+    It returns a :class:`MetDataset` containing the forecasted severity and/orenergy forcing.
```
thabbott:
```suggestion
    It returns a :class:`MetDataset` containing the forecast severity and/or energy forcing.
```

pycontrails/datalib/google_forecast.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Parameters
+    ----------
+    time : metsource.TimeInput | None
+        The time range for data retrieval, either a single datetime or (start, end) datetime range.
+        Input must be datetime-like or tuple of datetime-like
+        (`datetime`, :class:`pd.Timestamp`, :class:`np.datetime64`)
+        specifying the (start, end) of the date range, inclusive.
+    variables : metsource.VariableInput
+        Variable name (i.e. "contrails", "eef_per_m", ["contrails", "eef_per_m"])
```
thabbott:
```suggestion
        Variable name (i.e. "contrails", "eeef_per_m", ["contrails", "eeef_per_m"])
```

thabbott (commented):
Zeb asked me to take a quick look. I think this looks great--just a few minor quibbles related to naming.

@zebengberg, there are a few other chores we should take care of before we merge (updating the changelog, linking the demo notebook in the documentation). Do you want one of us to do them once Max's datalib work is finalized?

pycontrails/datalib/google_forecast.py
max-vogler: Good suggestion, done!

zebengberg:
> @zebengberg, there are a few other chores we should take care of before we merge (updating the changelog, linking the demo notebook in the documentation). Do you want one of us to do them once Max's datalib work is finalized?

@thabbott Yep, thanks. I'll do these after we merge but before we release.

PR #371, opened 2026-02-03
Fix/cocip output

### LINKED ISSUE #55: Consistency in `Cocip` outputs
Presently, the `Cocip` model attaches different variables to the `source` and `contrail` attributes depending on if flight waypoints generate persistent contrails. These differences should be documented or be made consistent.

- Avoid unnecessary operations in the default evaluation mode
- Attach any extraneous variables by setting `params["verbose_outputs"]`
- Consider running `Cocip` in a "low memory" mode in which the `contrail_list` attribute isn't used (instead, only the `latest_contrail` is kept during model evaluation). This would very helpful for better memory management in the contrails API.

zebengberg:
Closes #55 

## Breaking Changes

- The `Cocip.contrail` attribute is now always set to a `pd.DataFrame` after `eval` is called, even when no contrails are formed. Previously, it remained `None` in early-exit scenarios. The empty DataFrame has standardized column names and dtypes matching the structure produced when contrails persist.
- The `Cocip` model now adds consistent output columns to `source` regardless of whether contrails are formed. When no contrails form, columns such as `sdr_mean`, `rsr_mean`, `olr_mean`, `rf_sw_mean`, `rf_lw_mean`, `rf_net_mean`, and intermediate SAC/downwash columns (`width`, `depth`, `rhi_1`, `rho_air_1`, `iwc_1`, `f_surv`, etc.) are filled with `nan` values. This ensures that the output schema is consistent across runs of `Cocip.eval` but may slightly increase memory usage in scenarios where no contrails are formed.
- Remove the `Cocip.contrail_dataset` attribute. Previously this was computed when `Cocip.eval` was run with the `verbose_outputs` flag enabled. If needed, users can recompute recompute this via `xr.Dataset.from_dataframe(cocip.contrail.set_index(["timestep", "waypoint"]))`.

## Tests

- [ ] CI tests pass

### DIFF
--- a/docs/notebooks/run-cocip-on-flight.ipynb
+++ b/docs/notebooks/run-cocip-on-flight.ipynb
@@ -1018,7 +1018,7 @@
     "The `cocip` instance contains information about the contrail created when `cocip.eval` is run.\n",
     "\n",
     "- `cocip.source` is the original input flight with additional variables added\n",
-    "- `cocip.contrail` is a `pandas.DataFrame` with the contrail waypoints"
+    "- `cocip.contrail` is a `pandas.DataFrame` with the contrail properties at each evolution time step"
    ]
   },
   {


docs/notebooks/run-cocip-on-flight.ipynb
```diff
@@ -1015,11 +1015,10 @@
...
    "metadata": {},
    "source": [
-    "The model class contains information about the contrail created:\n",
+    "The `cocip` instance contains information about the contrail created when `cocip.eval` is run.\n",
     "\n",
-    "- `cocip.source` the original input flight\n",
-    "- `cocip.contrail` will be defined as a `pandas` DataFrame if a contrail is created. \n",
-    "- `cocip.contrail_dataset` is the same data but formatted as an `xarray` Dataset."
+    "- `cocip.source` is the original input flight with additional variables added\n",
+    "- `cocip.contrail` is a `pandas.DataFrame` with the contrail waypoints"
```
thabbott: Minor language thing: a DataFrame "with the contrail waypoints" makes it sound like this just includes flight waypoints where a contrail formed. How about "with contrail segment properties"?

pycontrails/datalib/leo_utils/landsat_metadata.py
```diff
@@ -192,7 +192,7 @@ def parse_ephemeris_landsat(ang_content: str) -> pd.DataFrame:
     -------
     pd.DataFrame
         A :class:`pandas.DataFrame` containing the ephemeris track with columns:
-        - EPHEMERIS_TIME: Timestamps of the ephemeris data.
+        - EPHEMERIS_TIME: Timestamps of the ephemeris data with datetime64[ns]
```
thabbott:
```suggestion
        - EPHEMERIS_TIME: Timestamps of the ephemeris data as datetime64[ns]
```

tests/unit/test_cocip.py
thabbott, line 1: Can we add some tests to make sure the columns in `Cocip.contrail` and `Cocip.eval` output don't depend on whether or not contrails are formed?

zebengberg: It was painful, but I came up with a unit test for this cb615d36b0e011d798018528cfcf382c7d7e5358

docs/notebooks/run-cocip-on-flight.ipynb
zebengberg: 4d78af0536d3eb1385720d12fb88ff10d2b90212

PR #376, opened 2026-02-11
Add example to intersect and plot Google forecast for a flight.

max-vogler:
#### Internals

> Add example to intersect and plot Google forecast for a flight.

## Tests

- [X] QC test passes locally (`make test`)
- [ ] CI tests pass

## Reviewer

> @thabbott


max-vogler:
Thanks Zeb, that's good feedback. I updated everything, except for cell 2 I ultimately chose to load a hardcoded timestamp, not a relative timestamp. This makes the execution reproducible — the cell outputs won't change on every invocation.

PR #398, opened 2026-04-01
Expose convective velocity scale as a user-configurable parameter

thabbott:
## Changes

This was requested by a user to support some sensitivity studies.

The changes are mostly straightforward. The only complication (which is why I'm asking Roger to review as well) is that I found a bug in the vertical diffusivity calculation, which (incorrectly) diagnoses vertical diffusivity based on the ratio of the convective velocity scale to the Brunt-Vaisala frequency rather than the *square* of the convective velocity scale. This isn't really an issue when running CoCiP in its default configuration, since the default value of 0.01 is consistent with the *square* of the convective velocity scale suggested by [Schumann and Graf (2013)](https://agupubs.onlinelibrary.wiley.com/doi/10.1002/jgrd.50184). It does produce changes when diagnosing the convective velocity scale based contrail radiative heating, though, because the value returned by [`radiative_heating.convective_velocity_scale`](https://github.com/contrailcirrus/pycontrails/blob/2da8cd34a9487ce7b4dfd4807805069ca14e6e8f/pycontrails/models/cocip/radiative_heating.py#L100) is used without squaring.

### Breaking changes

- Fix a bug in the vertical diffusivity calculation. Vertical diffusivity is now correctly diagnosed based on the square of the convective velocity scale. This changes default behavior only when the convective velocity scale is itself diagnosed based on differential radiative heating at the top and bottom of the contrail (`radiative_heating_effects = True`).

### Features

- Make the convective velocity scale used to calculate vertical diffusivity (similar to `w_N'` in Equation 35 of Schumann et al 2012) a user-configurable parameter.

## Tests

- [x] CI tests pass

## Reviewer

> @zebengberg, @roger-teoh


thabbott:
@zebengberg: I'm not going to update the failing `test_radiative_heating_effects_param` test until we can talk to Roger about whether the breaking change is likely to cause problems (for him or other users).

zebengberg (approved):
Thanks @thabbott ! This looks really clean.

PR #399, opened 2026-04-07
Feature/payload updates

roger-teoh:
Closes #XX

## Changes

- Improved approach to estimate payload mass for passenger aircraft and dedicated freighters, evaluated against the Brazillian ANAC dataset
- Integrate Dray et al. (2024) cargo load factor database (https://doi.org/10.1016/j.jairtraman.2024.102692), which provides the 2019 annual mean origin-destination specific cargo load factors in the passenger hold and dedicated freighters
- Discontinue use of the IATA cargo load factor database, as it does not break down between freight carried in the passenger hold and dedicated freighters
- Provide reference values for the number of seats for each passenger aircraft type to improve payload estimates. 

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer


roger-teoh:
@zebengberg Core functions have been written up in jet.py, but needs to be integrated correctly into aircraft_performance.py for the PS model and BADA workflows.

PR #405, opened 2026-04-30
Fix/ap fields

zebengberg:
### Fixes

- No longer set `Flight.attrs` with NaN or NaT values in `ChAviation` datalib when the metadata is missing.

### Internals

- Handle timezone-aware timestamps in met datalibs.

### Breaking changes

- Expose the maximum Mach number buffer applied in `PSFlight` via a new `AircraftPerformanceParams.max_mach_buffer` parameter. Previously this was hard-coded to 0.02. The default value has now changed from 0.02 to 0.0. With this new default, downstream users should expect a small change in `PSFlight` fuel flow estimates (typically a decrease) when working with noisy ADS-B. Set this to 0.02 to retain previous behavior.
- Change the `fl.attrs["aircraft_performance_model"]` value from `"PSFlight"` to `"PS"` when the `PSFlight` model is used for consistency with `BADAFlight`.
- Rename the function arguments `max_takeoff_weight` -> `amass_mtow` and `operating_empty_weight` -> `amass_oew` in `jet.update_aircraft_mass` and `jet.initial_aircraft_mass`. This is to standardize the naming convention for aircraft mass parameters across pycontrails. Typically these functions are not called directly so the risk that downstream code is affected by this change is low.

## Tests

- [ ] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,22 +1,28 @@
 # Changelog
 
-## [Unreleased]
+## 0.63.0
 
 ### Features
 
-- Support extrapolation in model-level-to-pressure-level interpolation
-  (ml\_to\_pl)
+- Support extrapolation in model-level-to-pressure-level interpolation (`pycontrails.datalib.ecmwf.ml_to_pl`).
+
 ### Fixes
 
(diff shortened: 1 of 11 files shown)


thabbott (approved):
No concerns from me

PR #408, opened 2026-05-12
Met utils dims

geraedts-google:
#### Fixes

- `pycontrailsmdataset.met_utils.ml_to_pl` handles cases where dataset dimension
  order is different from variable dimension order. confusingly xarray does not do a good job keeping these in sync. transpose() for instance changes the variable order but not the dataset order. The ds.assign call in model_levels.ml_to_pl can also do weird things to the dataset dimension order

## Tests

- [X] QC test passes locally (`make test`)
- [X] CI tests pass

## Reviewer

@zebengberg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,10 @@
 
 ## 0.63.1 [Unreleased]
 
-### Features
-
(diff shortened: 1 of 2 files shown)


zebengberg:
Thanks @geraedts-google, I'll try take a look at this later today.

PR #414, opened 2026-06-17
Fix/release

zebengberg:
### Fixes

- Fix release workflow that prevented v0.63.2 wheels from being uploaded to PyPI.

### Breaking changes

- No longer build wheels for macOS x86_64. (macOS arm64 wheels will continue to be built and supported.)

### Internals

- Move cibuildwheel configs from the `.github/workflows/release.yaml` workflow to the `pyproject.toml` for easier local testing.
- Remove mypy check for python 3.11 in the test workflow (numpy 2.5 only supports python 3.12 and above).
- Remove the `pytest.mark.unreliable` marker in favor of skipping tests that make network calls with `pytest.mark.skipif(OFFLINE)`. Set the `PYCONTRAILS_OFFLINE` environment variable to force these tests to be skipped regardless of network availability. This variable is now set in the release workflow to skip all tests that make network calls.
- Remove the `IN_GITHUB_ACTIONS` flag from the LEO tests, which was redundant with the `PYCONTRAILS_SKIP_LEO_TESTS` environment variable.

## Tests

- [x] CI tests pass

## Reviewer

> Select @ reviewer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,15 +214,7 @@ nb-test: ensure-era5-cached nb-clean-check nb-format-check nb-check-links
 # https://github.com/jupyterlab/pytest-check-links
 nb-check-links:
 	python -m pytest --check-links \
-        --check-links-ignore "https://doi.org/10.1175/JAMC-D-11-0242.1" \
-		--check-links-ignore "https://doi.org/10.1021/acs.est.9b05608" \
-		--check-links-ignore "https://doi.org/10.1021/acs.est.2c05781" \
-		--check-links-ignore "https://doi.org/10.5194/acp-17-403-2017" \
-		--check-links-ignore "https://doi.org/10.5194/acp-2022-169" \
-		--check-links-ignore "https://doi.org/10.5194/gmd-5-543-2012" \
-		--check-links-ignore "https://doi.org/10.5194/egusphere-2024-1361" \
-		--check-links-ignore "https://doi.org/10.5194/egusphere-2025-3312" \
-		--check-links-ignore "https://doi.org/10.5194/acp-22-10919-2022" \
+        --check-links-ignore "https://doi.org/*" \
 		--check-links-ignore "https://gmd.copernicus.org/articles/5/543/2012/" \
 		--check-links-ignore "https://github.com/contrailcirrus/pycontrails-bada" \
 		--check-links-ignore "https://ourairports.com" \
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ yamllint:
 	yamllint -d "{extends: default, rules: {line-length: {max: 100}}}" .
 
 mypy:
+	python -c 'import numpy; assert tuple(int(x) for x in numpy.__version__.split(".", 2)) >= (2, 5), f"numpy >= 2.5 required, got {numpy.__version__}"'
 	mypy pycontrails
 
 pytest:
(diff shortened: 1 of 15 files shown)


tests/unit/test_dwd.py
```diff
@@ -44,6 +44,7 @@ def test_list_forecasts(domain: str, count: int) -> None:
     assert len(forecasts) == count
 
 
+@pytest.mark.unreliable
```
thabbott: Why is this test marked as unreliable? The idea behind `pytest.mark.unreliable` was to mark tests that rely on some upstream service that's outside our control, but I think an invalid domain gets caught without needing to make a connection to the DWD server. (I was able to get the error locally with wifi turned off.)

tests/unit/test_goes.py
```diff
@@ -147,6 +150,7 @@ def goes_data() -> xr.DataArray:
     return downloader.get("2023-09-15T15:34")
 
 
+@pytest.mark.unreliable
```
thabbott: Similarly, why mark the parallax correction tests as unreliable?

tests/unit/test_dwd.py
```diff
@@ -71,6 +72,7 @@ def test_list_forecast_steps_warning(domain: str) -> None:
     assert len(steps) == 0
 
 
+@pytest.mark.unreliable
```
thabbott: Same comment about unreliable marker applies here.

tests/unit/test_goes.py
```diff
@@ -232,6 +241,7 @@ def test_goes_19() -> None:
     assert da["band_id"].values.tolist() == [2]
 
 
+@pytest.mark.unreliable
```
thabbott: The unreliable marker makes sense here, but I think we should also skip this test if we're offline.

Makefile
```diff
@@ -216,6 +217,13 @@ nb-check-links:
         --check-links-ignore "https://doi.org/10.1175/JAMC-D-11-0242.1" \
 		--check-links-ignore "https://doi.org/10.1021/acs.est.9b05608" \
 		--check-links-ignore "https://doi.org/10.1021/acs.est.2c05781" \
+		--check-links-ignore "https://doi.org/10.5194/acp-17-403-2017" \
```
thabbott: `--check-links-ignore` accepts regular expressions (https://github.com/jupyterlab/pytest-check-links). Maybe worth just skipping checks for any doi.org links?

thabbott (commented):
It's not clear to me what the purpose of some of the new `pytest.mark.unreliable` marks is. I think my intent when I added the mark was to add a way to skip tests that rely on some unreliable external service (mostly for the sake of avoiding failures during releases if a service happens to be doing something funny). So in general I know what to expect when I see
```python
@pytest.mark.unreliable
@pytest.mark.skipif(OFFLINE, reason="offline")
def test_something() -> None:
    ...
```
(this test needs an internet connection, and the service it accesses over the internet is unreliable enough that we want to ignore presumably-transient failures during releases) and when I see
```python
@pytest.mark.skipif(OFFLINE, reason="offline")
def test_something_else() -> None:
    ...
```
(this test needs an internet connection, but as long as we have an internet connection we always want to know if the test fails), but I'm not sure what the use case is for
```python
@pytest.mark.unreliable
def test_a_third_thing() -> None:
    ...
```

Broadly, I think our tests could fail unexpectedly because of issues with external services and external dependencies. The places where we interact with external services are pretty limited, so I think it makes sense to explicitly allow some failures during releases. In contrast we rely on external dependencies everywhere, so I don't think we have much ability to predict where they're likely to cause unexpected failures and probably shouldn't get in the habit of trying to mark those places (if that's even the purpose of the third type of mark).

tests/unit/test_dwd.py
zebengberg: Thanks, this one was unnecessary

zebengberg:
Thanks @thabbott !

I removed the unreliable markers and instead extended the `OFFLINE` check so that tests requiring network access don't run as part of the release workflow. 803253927e7803cf08daa483c8a60d4b279696b3

PR #427, opened 2026-08-25
Update OpenSky data access instructions

masorx:
Impala has been discontinued two years ago - there's a more modern Trino interface now.


zebengberg:
Thanks @masorx!

PR #428, opened 2026-08-25
Add Met Office UM datalib for cruise-level forecast data

JayGreen7:
## Changes

Adds `pycontrails.datalib.metoffice`, a new datalib for the UK Met Office global-deterministic-10km forecast product. There are some accompanying pieces to process the result into a MetDataset.

#### Internals

- Adds unit test coverage in `tests/unit/test_metoffice.py`.
- Adds the `metoffice` optional-dependency extra (`boto3`, `h5netcdf`, `h5py`, `platformdirs`).
- I had to pin `pytest<9` in dev, since `pytest-check-links` doesn't yet support pytest 9 ([jupyterlab/pytest-check-links#143](https://github.com/jupyterlab/pytest-check-links/pull/143) is open but unmerged). Unsure if this is a known issue.

## Tests

- [ ] CI tests pass

## Reviewers

@thabbott


zebengberg:
Thanks for this @JayGreen7!

I pulled down your branch to try it. I'm having difficult getting it finish in a timely manner. For example, the snippet below eventually finishes, but it takes about 10 minutes (both locally and on a google cloud vm)

```python
import pycontrails.datalib.metoffice as mo

moum = mo.MetOfficeUM("2026-01-01")
met = moum.open_metdataset()
```

Downloading the relevant netcdf file directly with `s3fs` or the aws cli only takes a few seconds

```sh
aws s3 cp --no-sign-request s3://met-office-atmospheric-model-data/global-deterministic-10km/20260101T0000Z/20260101T0000Z-PT0000H00M-temperature_on_pressure_levels.nc .
aws s3 cp --no-sign-request s3://met-office-atmospheric-model-data/global-deterministic-10km/20260101T0000Z/20260101T0000Z-PT0000H00M-relative_humidity_on_pressure_levels.nc .
```

The main bottleneck seems to be `h5py` calling `_S3RangeFile.read` thousands of times to download small chunks.

zebengberg:
@JayGreen7 Separately, do you know if the UK Met Office provides some flavor of cloud ice water content data? It's needed to compute tau cirrus within CoCiP.

JayGreen7:
> @JayGreen7 Separately, do you know if the UK Met Office provides some flavor of cloud ice water content data? It's needed to compute tau cirrus within CoCiP.

AFAIK, they do not publish any cloud ice water content so this datalib cannot support full CoCiP as is.

JayGreen7:
> Thanks for this @JayGreen7!
> 
> I pulled down your branch to try it. I'm having difficult getting it finish in a timely manner. For example, the snippet below eventually finishes, but it takes about 10 minutes (both locally and on a google cloud vm)
> 
> ```python
> import pycontrails.datalib.metoffice as mo
> 
> moum = mo.MetOfficeUM("2026-01-01")
> met = moum.open_metdataset()
> ```
> 
> Downloading the relevant netcdf file directly with `s3fs` or the aws cli only takes a few seconds
> 
> ```shell
> aws s3 cp --no-sign-request s3://met-office-atmospheric-model-data/global-deterministic-10km/20260101T0000Z/20260101T0000Z-PT0000H00M-temperature_on_pressure_levels.nc .
> aws s3 cp --no-sign-request s3://met-office-atmospheric-model-data/global-deterministic-10km/20260101T0000Z/20260101T0000Z-PT0000H00M-relative_humidity_on_pressure_levels.nc .
> ```
> 
> The main bottleneck seems to be `h5py` calling `_S3RangeFile.read` thousands of times to download small chunks.

Thanks for testing this @zebengberg. The 10-minute case you identified is 2026-01-01 which falls before a schema change the MetOffice made (around 2026-01-22). Pre-break files are way more finely chunked, which means my current approach is much slower for those. From my testing, post-break files are performant - perhaps you could test 2026-02-01?

zebengberg:
> From my testing, post-break files are performant - perhaps you could test 2026-02-01?

I tested on 2026-02-01 and a more recent date as well. I'm still seeing 4394 calls to `_S3RangeFile.read` (so similar to 2026-01-01).

JayGreen7:
> > From my testing, post-break files are performant - perhaps you could test 2026-02-01?
> 
> I tested on 2026-02-01 and a more recent date as well. I'm still seeing 4394 calls to `_S3RangeFile.read` (so similar to 2026-01-01).

You're right, my bad. For anything close to a global or all-levels fetch, whole-file download makes a lot more sense - I will push up a fix to switch to that. The byte-range approach was built against a narrower case (small region, fewer levels, where byte range and download perform similarly), which doesn't reflect what the datalib should support generally.

zebengberg:
@JayGreen7 sounds good!

If you do restructure to download the full netcdf, I'd suggest using `s3fs` / `fsspec` instead of `boto3` / `botocore` for your s3 handler. We have a lot of existing code in `pycontrails` which uses the former and would prefer to standardize. For an example, see the [Himawari datalib](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/himawari/himawari.py).

zebengberg:
> @JayGreen7 sounds good!
> 
> If you do restructure to download the full netcdf, I'd suggest using `s3fs` / `fsspec` instead of `boto3` / `botocore` for your s3 handler. We have a lot of existing code in `pycontrails` which uses the former and would prefer to standardize. For an example, see the [Himawari datalib](https://github.com/contrailcirrus/pycontrails/blob/main/pycontrails/datalib/himawari/himawari.py).

Here's one example of moving from `boto3` -> `s3fs`

c57950d9b7e2767

PR #429, opened 2026-08-27
Make GCPCacheStore picklable

JayGreen7:
## Changes

Fix `GCPCacheStore` so it no longer holds an unpicklable `google.cloud.storage.Client`/`Bucket` handle, allowing instances to be pickled and sent across process boundaries (i.e. to Apache Beam workers).

#### Internals

Previously, `GCPCacheStore` held a live `google.cloud.storage.Client` internally. Beam pickles everything, but `google.cloud.client.Client` cannot be pickled. If a `GCPCacheStore` is constructed and passed into Beam, it crashes. This fixes this behaviour by constructing a fresh property which handles the client separately.

## Tests

- [ ] CI tests pass

## Reviewer

@thabbott


thabbott:
@zebengberg: do you use `GCPCacheStore` for anything? If so, let me know if this change is going to cause issues for you. If not I'll merge it shortly.

zebengberg:
> @zebengberg: do you use `GCPCacheStore` for anything? If so, let me know if this change is going to cause issues for you. If not I'll merge it shortly.

Not using it, go ahead.

pycontrails/core/cache.py
```diff
@@ -551,18 +541,28 @@ def __repr__(self) -> str:
...
         -------
         :class:`google.cloud.storage.Client`
             Handle to Google Cloud Storage client
         """
-        return self._client
+        from google.cloud import storage
+
+        return storage.Client(project=self.project)
+
+    def _get_bucket(self) -> google.cloud.storage.Bucket:
```
thabbott:
```suggestion
    @property
    def _bucket(self) -> google.cloud.storage.Bucket:
```

pycontrails/core/cache.py
```diff
@@ -551,18 +541,28 @@ def __repr__(self) -> str:
...
+        """Return a handle to the GCS bucket, constructed from a fresh client."""
+        return self.client.bucket(self.bucket)
 
     @property
     @override
     def size(self) -> float:
         # get list of blobs below this path
-        blobs = self._bucket.list_blobs(prefix=self.cache_dir)
+        blobs = self._get_bucket().list_blobs(prefix=self.cache_dir)
```
thabbott:
```suggestion
        blobs = self._bucket().list_blobs(prefix=self.cache_dir)
```

pycontrails/core/cache.py
```diff
@@ -619,7 +619,7 @@ def exists(self, cache_path: str) -> bool:
             return True
 
         bucket_path = self.path(cache_path)
-        blob = self._bucket.blob(bucket_path)
+        blob = self._get_bucket().blob(bucket_path)
```
thabbott:
```suggestion
        blob = self._bucket().blob(bucket_path)
```

pycontrails/core/cache.py
```diff
@@ -678,7 +678,7 @@ def put(self, data_path: str | pathlib.Path, cache_path: str | None = None) -> s
         # get bucket and disk paths and blob
         bucket_path = self.path(cache_path)
         disk_path = self._disk_cache.path(cache_path)
-        blob = self._bucket.blob(bucket_path)
+        blob = self._get_bucket().blob(bucket_path)
```
thabbott:
```suggestion
        blob = self._bucket().blob(bucket_path)
```

pycontrails/core/cache.py
```diff
@@ -739,7 +739,7 @@ def get(self, cache_path: str) -> str:
         bucket_path = self.path(cache_path)
         disk_path = self._disk_cache.path(cache_path)
 
-        blob = self._bucket.blob(bucket_path)
+        blob = self._get_bucket().blob(bucket_path)
```
thabbott:
```suggestion
        blob = self._bucket().blob(bucket_path)
```

pycontrails/core/cache.py
```diff
@@ -807,7 +807,7 @@ def _dangerous_clear(self, confirm: bool = False, cache_path: str = "") -> None:
         self.clear_disk()
 
         # get list of blobs below this path
-        blobs = self._bucket.list_blobs(prefix=bucket_path)
+        blobs = self._get_bucket().list_blobs(prefix=bucket_path)
```
thabbott:
```suggestion
        blobs = self._bucket().list_blobs(prefix=bucket_path)
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Update the [ICAO EDB](https://www.easa.europa.eu/en/domains/environment/icao-aircraft-engine-emissions-databank) static files used in pycontrails emissions modeling from v31 to v32. With this update, pycontrails now supports emissions modeling for 26 new engines, bringing the total up to 884 engines (up from 858 engines in v31). This update includes new data for LEAP-1A, LEAP-1B, LEAP-1C and Trent 1000-TEN engines.
 - Update `emissions.py` to automatically use the latest `engine_uid` if the provided `engine_uid` has been superseded. (Supersession chains are followed to the end.)
 - Support parsing the `<Function>` elements of `<Waypoint>`s of the ARINC 633 OFP. An additional `"functions"` field is now included in the `Flight` output of `flightplan.parse_ofp_xml`.
+- Fix `GCPCacheStore` so it no longer holds an unpicklable `google.cloud.storage.Client`/`Bucket` handle, allowing instances to be pickled and sent across process boundaries (i.e. to Apache Beam workers).
```
thabbott:
```suggestion
- Update `GCPCacheStore` so it no longer holds an unpicklable `google.cloud.storage.Client`/`Bucket` handle, allowing instances to be pickled and sent across process boundaries (i.e. to Apache Beam workers).
```

thabbott (changes requested):
This mostly looks good. My only concern is that there's a pretty significant overhead (~250 ms) to creating a new `Client`, so it'd be nice to re-use a `Client` instance when pickleability isn't required. You might be able to do this by modifying some slots so they're `Client | None` or `Bucket | None` and adding a `pickleable` parameter that controls whether those slots are used to store reusable objects. Or, maybe this is a reason to prefer your original `__getstate__`/`__setstate__` solution. 

(I realize that removing the `_client` slot completely was my idea--some benchmarking I did since then makes me think it probably wasn't a good one.)

JayGreen7:
> This mostly looks good. My only concern is that there's a pretty significant overhead (~250 ms) to creating a new `Client`, so it'd be nice to re-use a `Client` instance when pickleability isn't required. You might be able to do this by modifying some slots so they're `Client | None` or `Bucket | None` and adding a `pickleable` parameter that controls whether those slots are used to store reusable objects. Or, maybe this is a reason to prefer your original `__getstate__`/`__setstate__` solution.
> 
> (I realize that removing the `_client` slot completely was my idea--some benchmarking I did since then makes me think it probably wasn't a good one.)

I understand the concern. I have added another fix so that `Client` construction happens once per `GCPCacheStore` instead of once per access. I've tested this and subsequent accesses take 0ms. 

The initial access takes longer than 250ms for me, but I suspect you may be using an SA and some network auth on my end is slowing things down.

thabbott:
This looks great--thanks!

My only remaining question is about the `google.cloud` import in the constructor. Is there a reason you want to leave it there despite it being unused until the `client` method's called? If not I'd suggest removing the import in the constructor and raising a `ModuleNotFoundError` in `client` instead.

JayGreen7:
> This looks great--thanks!
> 
> My only remaining question is about the `google.cloud` import in the constructor. Is there a reason you want to leave it there despite it being unused until the `client` method's called? If not I'd suggest removing the import in the constructor and raising a `ModuleNotFoundError` in `client` instead.

Agreed, just an oversight on my end. Pushed up a change.

thabbott:
Thanks. Everything looks good to me!

I'm going to merge temporarily to a different branch in contrailcirrus and open a new PR so I can run the checks, and assuming they all pass will merge that branch into main.

